PR #2, opened 2021-03-14
Add Markup View (with only markdown support)

sebastiantoh:
## Changes Made:

1. Add a very primitive web view (wrapper around WKWebView) that can only display raw html supplied.

2. Add an external library, Ink (https://github.com/JohnSundell/Ink), using Swift Package Manager (as opposed to CocoaPods). Purpose of using the library is to parse the given markdown text into HTML. 

3. Instead of having a `CardView` as per our initial planning, I've decided to do away with it since structures can't really be inherited. If I'm not wrong, the purpose of the `CardView` was solely for the styling. As such, I've added an extension to `View` that can apply some very basic styling (drop shadows) to the current view.

## Testing this PR:
1. Open `ContentView` and add `MarkupView()` to the body.
2. Build, run and start typing some markdown.

image


Dynavity/Dynavity/view/MarkupView.swift
```diff
@@ -0,0 +1,26 @@
...
+import SwiftUI
+
+struct MarkupView: View {
+    @ObservedObject var viewModel = MarkupViewModel()
+
+    var body: some View {
+        HStack {
+            TextEditor(text: $viewModel.markupTextBlock.rawText)
+            Divider()
+            WebView(rawHtml: viewModel.markupTextBlock.toHtml())
```
ianyong, line 17: Ideally, the `WebView` should only be scrollable if there is a need to and the rendered markup should be of the same size as the raw text. I think it might be easier to address this later on though; after we add the ability to add and resize canvas elements so that testing is easier.

ianyong (approved):
LGTM!

Dynavity/Dynavity/view/MarkupView.swift
sebastiantoh: Yep, thanks for pointing them out. Will keep them in mind!

PR #3, opened 2021-03-14
Add canvas and test elements

hanstirtaputra:
image

Have yet to implement the pan and zoom features of the canvas. But I will make a PR first so that we can start work on the other canvas elements!

I decided to use a ZStack to implement the Canvas, drawing inspiration from [this](https://www.raywenderlich.com/7705231-creating-a-mind-map-ui-in-swiftui) article. I think the main Canvas can have a ZStack and hold an `CanvasElementsMapView`, and later on a `EdgesMapView` when we implement the graph network feature.


Dynavity/Dynavity/model/CanvasElementProtocol.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import CoreGraphics
+
+protocol CanvasElementProtocol {
+    var id: UUID { get set }
+    var position: CGPoint { get set }
+    var text: String { get set }
+    var visualID: String { get }
```
sebastiantoh, line 8: What's the difference between visualID and id? (or rather, why is there a need for a visuaIID?)

Dynavity/Dynavity/model/CanvasElementProtocol.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import CoreGraphics
+
+protocol CanvasElementProtocol {
```
sebastiantoh, line 4: I think this can also be `protocol CanvasElementProtocol: Identifiable`? So we don't have to manually conform to Identifiable in the individual canvas elements

sebastiantoh (dismissed):
LGTM! One thing I'm not too sure about is that right now, your views are communicating directly with your models. Should it go through ViewModel instead? @ianyong any thoughts?

Btw, nice mindmap tutorial! I think it really fits our needs

ianyong:
> LGTM! One thing I'm not too sure about is that right now, your views are communicating directly with your models. Should it go through ViewModel instead? @ianyong any thoughts?
> 
> Btw, nice mindmap tutorial! I think it really fits our needs

Yea, I think we should have a ViewModel in between. But I think either @wltan or I can add it in since it's more related to the adding/resizing/rotating/toolbar functionalities. We can get this PR merged first so that the other tasks are unblocked.

Dynavity/Dynavity/model/CanvasElementProtocol.swift
hanstirtaputra: Yep this makes sense. Will change it, thanks!

sebastiantoh (dismissed):

Dynavity/Dynavity/model/CanvasElementProtocol.swift
hanstirtaputra: `visualID` here uses the UUID and text for this canvas element to create a unique ID, you're right it's not totally necessary since `UUID()` randomly generates a UUID with 4 random bytes. Was just a precaution! 

I think WL or Ian can change this as they deem fit for the actual canvas elements.

Dynavity/Dynavity/view/CanvasElementView.swift
```diff
@@ -0,0 +1,29 @@
...
+//  CanvasElementView.swift
+//  Dynavity
+//
+//  Created by Hans Sebastian Tirtaputra on 15/3/21.
+//
+
+import SwiftUI
+
+struct CanvasElementView: View {
+    static let width = CGFloat(150) // arbitrary value for now
```
ianyong: Could we mark this as a `TODO` so that it is obvious that this should be changed?

Dynavity/Dynavity/model/CanvasElementProtocol.swift
sebastiantoh: I see, thanks!

Dynavity/Dynavity/view/CanvasElementMapView.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+
+import SwiftUI
+
+struct CanvasElementMapView: View {
+    @Binding var elements: [CanvasElementProtocol]
+
+    var body: some View {
+        ZStack {
+            ForEach(elements, id: \.visualID) { element in
```
ianyong, line 15: Building onto @sebastiantoh's point, I think `id` can be used here instead of `visualID`? `visualID` seems redundant.

ianyong (changes requested):

Dynavity/Dynavity/view/CanvasElementMapView.swift
ianyong: Oops, just saw your [explanation](https://github.com/Dynavity/dynavity/pull/3#discussion_r594043289). However, my concern is that when the user edits the text of the canvas element, wouldn't the `visualID` change as well? This might cause issues. In addition, not all canvas elements will have text.

Anyway you're right, either @wltan or I can change this behaviour when we work on implementing the various canvas elements.

Dynavity/Dynavity/view/CanvasElementView.swift
hanstirtaputra: Got it, thanks!

ianyong (approved):
LGTM!

Dynavity/Dynavity/model/CanvasElementProtocol.swift
sebastiantoh: Nvm, I realise this doesn't work

PR #4, opened 2021-03-15
Add canvas interaction

hanstirtaputra:
## Changes Made:
1. Add canvas view model to remove direct dependency of `CanvasView` to `Canvas`
2. Updated project bundle identifier in order to run Dynavity on my iPad instead of Xcode simulator
3. Add pan and zoom gestures for canvas (zoom uses pinch gesture)


Dynavity/Dynavity/view/canvas/CanvasView.swift
```diff
@@ -1,21 +1,97 @@
...
 struct CanvasView_Previews: PreviewProvider {
     static var previews: some View {
-        let canvas = Canvas()
+        let viewModel = CanvasViewModel()
+        CanvasView(viewModel: viewModel)
+    }
+}
+
+extension CanvasView {
+    func handleDragChange(_ value: DragGesture.Value, containerSize: CGSize) {
```
sebastiantoh: Is `containerSize` necessary? It doesn't seem to be used in this function

Dynavity/Dynavity/view/canvas/CanvasView.swift
```diff
@@ -1,21 +1,97 @@
...
+        )
+    }
+
+    func scaledOffset(_ scale: CGFloat, initialValue: CGPoint) -> CGPoint {
+        let newX = initialValue.x * scale
+        let newY = initialValue.y * scale
+        return CGPoint(x: newX, y: newY)
+    }
+
+    func clampedScale(_ scale: CGFloat, initialValue: CGFloat?) -> (scale: CGFloat, didClamp: Bool) {
```
sebastiantoh, line 79: I think it may be good to have a generic `clamp` function since it'll likely be used by other resizing / scaling operations as well. The implementation here https://stackoverflow.com/a/62044583 lgtm

But I guess this can be done later on when we actually do use a `clamp` function elsewhere.

Dynavity/Dynavity/view/canvas/CanvasView.swift
```diff
@@ -1,21 +1,97 @@
 import SwiftUI
 
 struct CanvasView: View {
-    @ObservedObject var canvas: Canvas
+    @ObservedObject var viewModel = CanvasViewModel()
+
+    // For drag interactions
+    @State var portalPosition = CGPoint()
```
sebastiantoh: Can this be renamed to `mapViewPositionFromCentre` or something else if it's too long? I think it can be quite hard to understand what `portal` refers to

Dynavity/Dynavity/view/canvas/CanvasView.swift
```diff
@@ -1,21 +1,97 @@
...
+    @State var dragOffset = CGSize()
+    @State var isDragging = false
+
+    // For zoom interactions
+    @State var zoomScale = CGFloat(1.0)
+    @State var initialZoomScale: CGFloat?
+    @State var initialPortalPosition: CGPoint?
```
sebastiantoh: Likewise for this

sebastiantoh (changes requested):
LGTM, tested and it works well! Just some minor nitpicks

Dynavity/Dynavity/view/canvas/CanvasView.swift
```diff
@@ -1,21 +1,97 @@
...
+        GeometryReader { geometry in
+            ZStack {
+                Rectangle()
+                CanvasElementMapView(elements: $viewModel.canvas.canvasElements)
+                    .scaleEffect(zoomScale)
+                    .offset(x: portalPosition.x + dragOffset.width,
+                            y: portalPosition.y + dragOffset.height)
+                    .animation(.easeIn)
+            }
+            .gesture(DragGesture()
```
ianyong: Could we extract the gestures out into their own variables? This will make `body` cleaner.

Dynavity/Dynavity.xcodeproj/project.pbxproj
```diff
@@ -531,17 +535,20 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
+				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_ASSET_PATHS = "\"Dynavity/Preview Content\"";
+				DEVELOPMENT_TEAM = STWSL7H5JX;
```
ianyong: Not too sure if we should be committing this change since everyone else will face conflicts using this config.

Dynavity/Dynavity/view/canvas/CanvasView.swift
hanstirtaputra: Right thanks for that! Initially this drag took tried to account for `CanvasElements` as well, but I'll remove it since it's not necessary anymore

Dynavity/Dynavity/view/canvas/CanvasView.swift
hanstirtaputra: Yep good call! Will create a generic `clamp` function for `Comparable` as shown in the implementation linked

Dynavity/Dynavity.xcodeproj/project.pbxproj
hanstirtaputra: Hmm alright, will try to find a workaround for this, otherwise each time I will have to change the team name before the last commit since I'll need to change the DEVELOPMENT_TEAM to run it on my iPad

sebastiantoh (approved):

ianyong (approved):

PR #5, opened 2021-03-16
LaTeX + Plain Text support + Colours 

sebastiantoh:
# Changes
1. Add rendering of LaTeX text. This is done by sending a POST request to an external API (basically a Pandoc wrapper), which responds with the HTML. Given that Pandoc supports conversion of Markdown to HTML as well, I've updated the rendering of Markdown to use a similar implementation. There's also a slight delay introduced between each requests. As such, the rendering may not be instantaneous.
  - Note: the support for LaTeX is quite limited. Some commands are not supported, and would fail silently. Comments using `%` are not supported as well. For invalid syntax, the WebView would display an error instead
2. Remove external dependency on `Ink` (previously used for its Markdown parser)
3. Add a bunch of different colours that can be used throughout the application
4. Add plaintext support

IMG_0010


Dynavity/Dynavity/view-model/MarkupViewModel.swift
```diff
@@ -1,12 +1,71 @@
...
+            let httpResponse = response as? HTTPURLResponse
+
+            guard httpResponse?.statusCode != 400,
+                  let rawHtml = String(bytes: data, encoding: stringEncoding) else {
+                return "Error: Invalid / unsupported syntax used."
+            }
+
+            return rawHtml
+        }
+        .assertNoFailure() // only source of failure is due to invalid URL, which should have been validated
```
ianyong, line 47: Should we try to have comments on a separate line? Preferably the line before?

ianyong (commented):
I have some concerns over being able to switch between the various markup languages and plaintext freely as I don't think it makes too much sense to be able to do so. I feel that whether the text block is plaintext/markdown/LaTeX/code should be decided at the point of creation. This can be handled through the use of context menus which I feel will be more intuitive to the user. See #6 for an example of what I mean.

Dynavity/Dynavity/model/MarkupTextBlock.swift
```diff
@@ -1,28 +1,33 @@
-//
-//  MarkupTextBlock.swift
-//  Dynavity
-//
-//  Created by Sebastian on 14/3/21.
-//
```
ianyong, line 6: I support this change! Removed header comments from my PR as well.

hanstirtaputra (dismissed):
LGTM

sebastiantoh:
> I have some concerns over being able to switch between the various markup languages and plaintext freely as I don't think it makes too much sense to be able to do so. I feel that whether the text block is plaintext/markdown/LaTeX/code should be decided at the point of creation. This can be handled through the use of context menus which I feel will be more intuitive to the user. See #6 for an example of what I mean.

Sure, I'll remove it!

sebastiantoh:
Closing in favour of #7 because GitHub doesn't seem to be properly showing the diff

PR #6, opened 2021-03-16
Add toolbar

ianyong:
* Add toolbar
  * Currently only has an `Add` button that brings up a context menu:
  image
  * Let me know if you think any icons should be changed.
* Fix canvas elements overlapping the status bar (and by extension, the toolbar).
* Fix canvas being black in light mode.


hanstirtaputra (approved):
LGTM, I think the icons are fine!

PR #7, opened 2021-03-17
 LaTeX + Plain Text support + Colours 

sebastiantoh:
# Changes
1. Add rendering of LaTeX text. This is done by sending a POST request to an external API (basically a Pandoc wrapper), which responds with the HTML. Given that Pandoc supports conversion of Markdown to HTML as well, I've updated the rendering of Markdown to use a similar implementation. There's also a slight delay introduced between each requests. As such, the rendering may not be instantaneous.
  - Note: the support for LaTeX is quite limited. Some commands are not supported, and would fail silently. Comments using `%` are not supported as well. For invalid syntax, the WebView would display an error instead
2. Remove external dependency on `Ink` (previously used for its Markdown parser)
3. Add a bunch of different colours that can be used throughout the application
4. Add plaintext support by having separate MVVM for it (P.S. @ianyong I recall you mention that prof said we shouldn't have 1 view, 1 model, 1 view-model for each entity?)

P.S. Please merge by squashing. For some reason, my commit history is messed up


ianyong:
> 3. Add plaintext support by having separate MVVM for it (P.S. @ianyong I recall you mention that prof said we shouldn't have 1 view, 1 model, 1 view-model for each entity?)

Right, the context for this was my complaint of having a lot of dependency injection in my views (since I had to pass my state into each and every view model I had). Prof Wai Kay's suggestion was that view models can be reused across views if they're similar enough in terms of functionality. This is especially true if view composition is heavily made use of, such that a single "feature" from the user's perspective is actually made up of multiple smaller views. In such a context, it might make more sense to have a single view model servicing the views that make up the "feature", rather than a separate view model for each view.

sebastiantoh:
> > 1. Add plaintext support by having separate MVVM for it (P.S. @ianyong I recall you mention that prof said we shouldn't have 1 view, 1 model, 1 view-model for each entity?)
> 
> Right, the context for this was my complaint of having a lot of dependency injection in my views (since I had to pass my state into each and every view model I had). Prof Wai Kay's suggestion was that view models can be reused across views if they're similar enough in terms of functionality. This is especially true if view composition is heavily made use of, such that a single "feature" from the user's perspective is actually made up of multiple smaller views. In such a context, it might make more sense to have a single view model servicing the views that make up the "feature", rather than a separate view model for each view.

I see... so in this case it's fine right? Since the functionalities of the textblock and markuptextblock are quite different? (Only commonality is displaying the text editor)

Dynavity/Dynavity/model/MarkupTextBlock.swift
```diff
@@ -1,28 +1,39 @@
...
+        }
+        set {
+            textBlock.text = newValue
+        }
+    }
+    var fontSize: CGFloat {
+        textBlock.fontSize
+    }
+
+    var textBlock = TextBlock()
```
ianyong: Would it be better to have the stored properties listed before the computed properties? I was a bit confused over where `textBlock` came from when I first read this part of the code.

ianyong (dismissed):
LGTM!

ianyong:
> > > 1. Add plaintext support by having separate MVVM for it (P.S. @ianyong I recall you mention that prof said we shouldn't have 1 view, 1 model, 1 view-model for each entity?)
> > 
> > 
> > Right, the context for this was my complaint of having a lot of dependency injection in my views (since I had to pass my state into each and every view model I had). Prof Wai Kay's suggestion was that view models can be reused across views if they're similar enough in terms of functionality. This is especially true if view composition is heavily made use of, such that a single "feature" from the user's perspective is actually made up of multiple smaller views. In such a context, it might make more sense to have a single view model servicing the views that make up the "feature", rather than a separate view model for each view.
> 
> I see... so in this case it's fine right? Since the functionalities of the textblock and markuptextblock are quite different? (Only commonality is displaying the text editor)

Yup, I agree with your assessment!

ianyong (approved):

PR #9, opened 2021-03-17
Add image canvas element

ianyong:
## Features

* Add image canvas elements.
  * Images can be either imported from photo gallery, or via camera.

## Additional Details

* Remove `visualID` from `CanvasElementProtocol`.
  * `visualID` is the concatenation of the `id` and `text` of the canvas element. However, not all canvas elements have text (such as image canvas elements).
  * The `id` of the canvas element itself is already almost always guaranteed to be unique.
  * See https://github.com/Dynavity/dynavity/pull/3#discussion_r594021486 for the previous discussion on this.
* Share `CanvasViewModel` with `ToolbarView`.
  * This is to allow for the adding of canvas elements through the toolbar.
* Update `info.plist` with `NSCameraUsageDescription` and `NSPhotoLibraryUsageDescription` so that the application can access these resources.

## Design Trade-off

Given that there are many different kinds of `CanvasElementProtocol`s, each with its own model and view, we need a way to implement polymorphic behaviour in our rendering of canvas element views. In particular, the `CanvasElementMapView` has to iterate over the array of `CanvasElementProtocol`s and render the correct view for each of the canvas elements.

Unfortunately, SwiftUI views are structs which do not have inheritance. As such, we are limited to the following two solutions:

1. Add a `toView` function to the `CanvasElementProtocol` which returns an `AnyView` (a type erased wrapper around a view which enables us to return multiple view types from a single function). Each canvas element model returns its own view (wrapped in an `AnyView`), when its `toView` function is called.
1. Use something akin to a switch statement within the `ForEach` loop that iterates over the array of `CanvasElementProtocol`s in `CanvasElementMapView`. For each canvas element, we check its type against all known canvas element types until we find a match.

In the end, I went with the second option as the first option has the following two glaring concerns:

1. The addition of the `toView` function results in tighter coupling between the model and the view in the MVVM architecture as the model then has a dependency on the view. This also defeats the entire purpose of using the MVVM architecture.
1. While the use of `AnyView` should not cause much issues within this specific context, it is still good practice to avoid the use of type erasure so that we can benefit from type safety.

In contrast, the second option does not have any of the issues highlighted above. Of course, there is a risk of ending up with a very large switch statement with the addition of more canvas elements.


Dynavity/Dynavity/view/canvas/elements/ImageCanvasElementView.swift
```diff
@@ -0,0 +1,17 @@
+import SwiftUI
+
+struct ImageCanvasElementView: View {
+    var imageCanvasElement: ImageCanvasElement
```
sebastiantoh: Should this go through a ViewModel instead? Otherwise View will be communicating directly with Model

Dynavity/Dynavity/view/CanvasElementMapView.swift
```diff
@@ -12,9 +12,16 @@ struct CanvasElementMapView: View {
 
     var body: some View {
         ZStack {
-            ForEach(elements, id: \.visualID) { element in
-                CanvasElementView(element: element)
-                    .offset(x: element.position.x, y: element.position.y)
+            ForEach(elements, id: \.id) { element in
+                // The group allows us to have common view modifiers.
```
sebastiantoh, line 16: By common modifier, you mean `.offset(x: element.position.x, y: element.position.y)`? i.e. you don't need to write `imageView.offset(...)` and  `elementView.offset(...)`?

Would it be better to use a similar construct like so: https://stackoverflow.com/a/34800922, then just write offset once on the assigned view?

Dynavity/Dynavity/view/ToolbarView.swift
```diff
@@ -51,6 +55,17 @@ struct ToolbarView: View {
         HStack {
             Spacer()
             addButton
+            // SwiftUI does not allow attaching multiple sheets to the same element.
```
sebastiantoh: Would it be better to use something like https://stackoverflow.com/questions/58837007/multiple-sheetispresented-doesnt-work-in-swiftui

i.e. use an enum to represent the currently displayed sheet. The enum can also contain the required arguments for the imagepicker (aka sourceType). 

Reason is because I'm thinking the current implementation may be susceptible to the buggy situation where multiple sheets are presented concurrently if state is poorly handled (which I'm assuming is non-ideal, at least for now)

sebastiantoh (dismissed):
LGTM. Tested and works well. Just some possible alternative implementations to consider

Dynavity/Dynavity/view/canvas/elements/ImageCanvasElementView.swift
ianyong: You're right, especially since we will be adding contextual actions to the various canvas elements later on. I'll make the change.

However, if this were just a dumb view (only displaying data from the model with no other functionalities), I'm inclined to not pass the model through a view model. While I acknowledge that this creates a dependency from the view to the model, wrapping the model in a view model (whose sole purpose is to contain the model) feels extremely redundant.

Dynavity/Dynavity/view/canvas/elements/ImageCanvasElementView.swift
```diff
@@ -0,0 +1,21 @@
...
+
+struct ImageCanvasElementView: View {
+    @ObservedObject var viewModel: ImageCanvasElementViewModel
+
+    init(imageCanvasElement: ImageCanvasElement) {
+        self.viewModel = ImageCanvasElementViewModel(imageCanvasElement: imageCanvasElement)
+    }
+
+    var body: some View {
+        Image(uiImage: viewModel.imageCanvasElement.image)
```
sebastiantoh, line 11: One more thing: do we want to add a drop shadow to the image as well? Using `addCardOverlay()`

Dynavity/Dynavity/view/canvas/elements/ImageCanvasElementView.swift
sebastiantoh: I see. Yep, makes sense to me!

Dynavity/Dynavity/view/CanvasElementMapView.swift
ianyong: > By common modifier, you mean `.offset(x: element.position.x, y: element.position.y)`? i.e. you don't need to write `imageView.offset(...)` and `elementView.offset(...)`?

Yup!

> Would it be better to use a similar construct like so: https://stackoverflow.com/a/34800922, then just write offset once on the assigned view?

Could you elaborate further on this? The `Group` is there because the Swift compiler is not smart enough to determine that the output of the `if`-`else` block is guaranteed to be a `View`. Not too sure how that solves this issue.

Dynavity/Dynavity/view/canvas/elements/ImageCanvasElementView.swift
ianyong: Right, I've applied the card overlay to all canvas elements in `CanvasElementMapView`.

Dynavity/Dynavity/view/CanvasElementMapView.swift
sebastiantoh: > Could you elaborate further on this? The Group is there because the Swift compiler is not smart enough to determine that the output of the if-else block is guaranteed to be a View. Not too sure how that solves this issue.

Whoops, you're right. My bad!

Dynavity/Dynavity/view/canvas/elements/ImageCanvasElementView.swift
sebastiantoh: Could you also remove all other instances where this `addCardOverlay()` is called? So we don't call it twice (not sure if it'll affect the appearance). Thanks!

ianyong: Done, thanks for pointing this out!

sebastiantoh (approved):

PR #10, opened 2021-03-18
Add name field to Canvas and add SideMenu

sebastiantoh:
Changes:
- Extract the code in `DynavityApp` into a separate view called `MainView`. This is in preparation for other views  which may be at the same level e.g. the home page / canvas selection view
  - P.S. Any suggestions on more descriptive name, rather than `MainView`?
- Shift Markup and Text canvas elements into the `elements` folder and remove remnant `visualID` in these elements as well.
- Add a `name` field to `Canvas` - this is to facilitate searching / filtering of canvases
- Add a SideMenu that displays additional info related to the particular canvas.
    - This SideMenu can be opened by clicking the rightmost button in the toolbar.
    - For now, this SideMenu can only be dismissed by creating a rightward dragging motion. I'm considering adding a dismiss on tap, but have decided against it for now, as it may complicate things esp if there are tap gestures in the SideMenu as well

IMG_0011


Dynavity/Dynavity/view/MainView.swift
```diff
@@ -0,0 +1,66 @@
+import SwiftUI
+
+struct MainView: View {
+    @ObservedObject private var canvasViewModel = CanvasViewModel()
```
hanstirtaputra, line 4: Should the `CanvasViewModel` be initialized in each `CanvasVIew` instead? Since we would likely have multiple canvases in a `MainView`, so each `CanvasView` would have their own `CanvasViewModel` instead of sharing a common one from `MainView`?

hanstirtaputra (approved):
LGTM! Just one question regarding the `CanvasViewModel`, though I think this is fine for now!

Dynavity/Dynavity/view/MainView.swift
sebastiantoh: hmm by `MainView`, I'm thinking of the view that the user would spend the most time on, which is the view where the user can modify the canvas. So `MainView` contains only 1 canvas, unless the intention is for the user to have multiple canvases open at the same time?

hanstirtaputra: I see! Alright if `MainView` contains only 1 canvas, then I think its good!

PR #11, opened 2021-03-18
Integrate textblocks with the Add button

sebastiantoh:
Changes:
1. Change `Canvas` into a struct
  - Rationale is because SwiftUI does not support nested ObservableObjects which results in the CanvasElementMap not re-rendering even though the underlying canvasElements has been updated.
  - Some references:
    - https://stackoverflow.com/questions/61280353/change-on-observableobject-in-published-array-does-not-update-the-view
    - https://stackoverflow.com/questions/58406287/how-to-tell-swiftui-views-to-bind-to-nested-observableobjects

2. Integrate text blocks with the Add button

Known Issues:
- Text does not persist for some reason (e.g. after dragging / zooming on the canvas, the text disappears). I'll revisit this issue some other time


ianyong (approved):
LGTM!

PR #12, opened 2021-03-19
Add Annotations

hanstirtaputra:
## Changes
1. Canvas implementation now uses a bridge between SwiftUI and UIKit. This is for compatibility with PencilKit's `PKCanvasView` which is required for annotations.
2. Pan gestures now use two fingers (instead of one finger).
3. Background colour is not set. Previously we used white background. This is so that the ` CanvasElementMapView` will not be covered by the white background color.

## Notes
- Annotation is not a `CanvasElementProtocol` as we initially discussed during our previous meeting. Instead it one giant `AnnotationCanvas` which stores the `PKDrawing` - a representation of all the annotations.
- The middle point of the canvas is currently at `(x: 250000, y: 250000)` instead of (0, 0). This allows the user to pan in all directions before hitting the end of the "infinite" canvas.
- To test this with the previously added `CanvasImageElement` feature in #9 , the position of the `CanvasElement` must be at (250000,250000) in order for it to be seen at the start without needing to pan.

## Design decisions:
- Changed the entire `CanvasView` to integrate it with `PKCanvasView` instead of overlaying an `AnnotationView` on the `CanvasView`. This is because the previous zoom implementation will result in the overlayed `AnnotationView` to be scaled as well, which prevents the user from annotating freely, anywhere on the canvas.

Test with Images-1

* Note that the screenshot was taken prior to merging #10 into my branch, but have tested after merging and it works!


Dynavity/Dynavity/view/canvas/CanvasView.swift
```diff
@@ -56,47 +22,104 @@ struct CanvasView_Previews: PreviewProvider {
...
+        populateAnnotationCanvasView()
     }
 
-    func handleScaleChange(_ value: CGFloat) {
-        if initialZoomScale == nil {
-            initialZoomScale = zoomScale
-            initialViewPositionFromCentre = viewPositionFromCentre
+    private func cleanAnnotationCanvasView() {
+        annotationCanvasView.subviews.forEach {
+            if $0.accessibilityLabel == "CanvasElements" {
```
sebastiantoh, line 74: Is it possible to remove the view by checking the type instead?

sebastiantoh (commented):
Is it supposed to work only with the Apple pencil? If I try drawing with my finger, it doesn't really recognise the gesture. Like I can only draw dots. 

Also, if I zoom out to the maximum, the red test squares will disappear then reappear momentarily after. 



https://user-images.githubusercontent.com/24363622/111776205-ae212800-88ec-11eb-879e-037d3bbe76bf.mp4

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
```diff
@@ -22,4 +24,9 @@ class CanvasViewModel: ObservableObject {
         // Reset the selected image.
         selectedImage = nil
     }
+
+    func storeAnnotation(_ drawing: PKDrawing) {
```
sebastiantoh, line 29: This function doesn't seem to be doing anything. Should `var` be removed?

Dynavity/Dynavity/view/canvas/CanvasView.swift
hanstirtaputra: So far, I've tried to check by type but the `CanvasElementMapView` is no longer a SwiftUI view, but converted to a `UIView` so that it can be added as a subview to the `AnnotationCanvasView`. So the `UIView` can not be checked if its a `CanvasElementMapView`, hence I added the accessibility label. Any suggestions for this?

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
hanstirtaputra: Whoops. You're right, I'll remove the `var`, thanks!

Dynavity/Dynavity/view/canvas/CanvasView.swift
sebastiantoh: I see. Then I think it's fine. Thanks!

hanstirtaputra:
> Also, if I zoom out to the maximum, the red test squares will disappear then reappear momentarily after.

Yep this is a known issue, ps didn't mention it in the initial comments. Still trying to find a fix for this! If it isn't too big an issue we can merge this in first to not block the other implementations while I find a fix?

sebastiantoh (approved):

sebastiantoh:
> > Also, if I zoom out to the maximum, the red test squares will disappear then reappear momentarily after.
> 
> Yep this is a known issue, ps didn't mention it in the initial comments. Still trying to find a fix for this! If it isn't too big an issue we can merge this in first to not block the other implementations while I find a fix?

Yep, ok!

PR #13, opened 2021-03-19
Add Home Page 

sebastiantoh:
Changes
- Add a Home Page, which contains a canvas selection view along with a search bar
- Filtering logic of canvases based on search term should be working, though it's currently only tested with a `String` instead of the actual `Canvas` model.
- A canvas in the Canvas Selection View is currently a document icon and its name. The document icon is likely to be replaced with some thumbnail of the actual canvas (probably requires the Storage implementation to be out first before this can be done)
- Tapping on any canvas in the Canvas Selection View will take you to the canvas editing view. There is currently no way to go back to home page once you have selected a canvas (i've disabled the back button in the NavigationView because it doesn't play well with our toolbar)

IMG_0014

### DIFF
--- a/Dynavity/Dynavity/DynavityApp.swift
+++ b/Dynavity/Dynavity/DynavityApp.swift
@@ -11,7 +11,7 @@ import SwiftUI
 struct DynavityApp: App {
     var body: some Scene {
         WindowGroup {
-            MainView()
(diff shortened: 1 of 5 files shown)


ianyong (approved):
LGTM!

PR #14, opened 2021-03-19
Fix zoom bug

hanstirtaputra:
## Changes Made
- Added a workaround that solves the disappearing and reappearing animation of the `CanvasElementMapView` elements when the user zoom's past the maximum or minimum zoom scale (issue highlighted in #12 ).


ianyong (approved):
LGTM! Tested, unable to replicate the issue as described in https://github.com/Dynavity/dynavity/pull/12#pullrequestreview-616315073.

PR #15, opened 2021-03-19
Add PDF canvas element

ianyong:
## Features

* Add PDF canvas elements.
  * PDF files are imported via a document picker.
  * PDF files are displayed using `PDFKit`'s `PDFView`.

## Additional Details

* Update `info.plist` with `NSDocumentsFolderUsageDescription` so that the application can access these resources.
* Add `canvasCenter` to `CGPoint` to address the change in origin in #12.

### Handling different behaviours upon dismissing the sheet

Previously, we decided to use an enum to represent the currently displayed sheet. See https://github.com/Dynavity/dynavity/pull/9#discussion_r595957187 for more information. However, we run into the issue where each sheet can only have a single `onDismiss` action. With the addition of PDF canvas elements, we now need to call either `addImageCanvasElement` or `addPdfCanvasElement` in `CanvasViewModel`, depending on which sheet was last displayed.

The first solution I tried was to create another function in `CanvasViewModel` that called either one of the above functions by checking the value of `activeSheet`. Unfortunately, this is not possible as `activeSheet` would have already been set to `nil` by the time `onDismiss` is called.

Subsequently, I moved the `activeSheet` variable out of `ToolbarView` and into `CanvasViewModel`. There, I added setter functions for `activeSheet` (i.e., `displayCamera`, `displayPhotoGallery`, `displayPdfPicker`). In addition, I also separately kept a private variable `lastActiveSheet`. This variable was set along with `activeSheet` in each of the setter functions. By doing so, we are able to know what the value of `activeSheet` was before the sheet was dismissed. This solution was not ideal for the following reasons:
1. Unable to enforce `private(set)` on the `activeSheet` variable in `CanvasViewModel`. This is because the sheet itself accesses `activeSheet` through a two-way binding and thus requires write access. Unfortunately, this means that we cannot enforce changing the value of `activeSheet` only through out setter functions.
1. The use of the `activeSheet` and `lastActiveSheet` variables is not very elegant as the states of both variables have to be constantly kept synchronised. Future extensions to the code can very easily break the invariants we have set out, making the code brittle.
1. I discovered a separate issue that renders this solution unfeasible anyway. See below for a detailed explanation of the issue.

### Delay in document picker setting the URL of the picked file

Unlike the image picker, the document picker has a slight delay in updating the `selectedFile` binding. With the previous solution discussed above in place, it was found that `addPdfCanvasElement` would trigger before the `selectedFile` binding was updated by the document picker. This might be due to the file picker having to create a copy of the file. However, this also raises the possibility that the image picker might suffer from this behaviour too (if the selected image is huge).

Clearly, some form of synchronisation is needed to solve this issue. Potential solutions include busy waiting, or setting a fixed delay before the functions to add the canvas elements are called. The former is undesirable as we will needlessly use up CPU cycles (as well as fail CS3217 since Prof Wai Kay will get an aneurysm if he sees us do that), while the latter not only makes the application feel less responsive, it also does not guarantee that the bindings will be updated before the functions are called.

Instead, we have a very elegant solution, which is to use the `didSet` property observer on our `selectedImage` and `selectedFile` bindings.

**Update:** @sebastiantoh correctly identified that `didSet` will trigger multiple times which is not ideal. After further consideration, I realised that I can just pass `addImageCanvasElement` and `addPdfCanvasElement` as callback functions to the image picker and document picker respectively. Not only does this solve the issue of the order of execution being unsynchronised, it also reduces coupling between `CanvasViewModel` and the respective pickers as we no longer need to maintain the `selectedImage` and `selectedFile` variables.

### Issues

When adding PDFs, they follow the dimensions of the last added canvas element. This should be resolved once the logic to place items on the canvas based off the current view of the canvas is added.

### DIFF
--- a/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
+++ b/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
@@ -9,6 +9,12 @@ import SwiftUI
 
 class CanvasViewModel: ObservableObject {
     @Published var canvas = Canvas()
+
+    // Sheets
+    @Published var activeSheet: ToolbarView.ActiveSheet?
+    private var lastActiveSheet: ToolbarView.ActiveSheet?
+
+    // Selected media
     @Published var selectedImage: UIImage?
     @Published var selectedFile: URL?
 
@@ -42,5 +48,36 @@ class CanvasViewModel: ObservableObject {
 
     func addMarkUpTextBlock(markupType: MarkupTextBlock.MarkupType) {
         canvas.addElement(MarkupTextBlock(markupType: markupType))
+    }
+}
+
+// MARK: Sheets
+extension CanvasViewModel {
+    func displayCamera() {
+        activeSheet = .camera
+        lastActiveSheet = .camera
+    }
+
+    func displayPhotoGallery() {
+        activeSheet = .photoGallery
+        lastActiveSheet = .photoGallery
+    }
+
+    func displayPdfPicker() {
+        activeSheet = .pdfPicker
+        lastActiveSheet = .pdfPicker
+    }
+
+    func handleSheetDismiss() {
+        switch lastActiveSheet {
+        case .camera:
+            addImageCanvasElement()
+        case .photoGallery:
+            addImageCanvasElement()
+        case .pdfPicker:
+            addPdfCanvasElement()
+        default:
+            assert(false, "When `handleSheetDismiss` is called, `lastActiveSheet` should not be `nil`.")
+        }
     }
 }
--- a/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
+++ b/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
@@ -10,6 +10,7 @@ import SwiftUI
 class CanvasViewModel: ObservableObject {
     @Published var canvas = Canvas()
     @Published var selectedImage: UIImage?
+    @Published var selectedFile: URL?
 
     func addImageCanvasElement() {
         guard let image = selectedImage else {
--- a/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
+++ b/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
@@ -5,40 +5,14 @@ class CanvasViewModel: ObservableObject {
     @Published var canvas = Canvas()
     @Published var annotationCanvas = AnnotationCanvas()
 
-    // Selected media
-    @Published var selectedImage: UIImage? {
-        didSet {
-            addImageCanvasElement()
-        }
-    }
-    @Published var selectedFile: URL? {
-        didSet {
-            addPdfCanvasElement()
-        }
-    }
-
-    func addImageCanvasElement() {
-        guard let image = selectedImage else {
-            return
-        }
-
+    func addImageCanvasElement(from image: UIImage) {
         let imageCanvasElement = ImageCanvasElement(image: image)
         canvas.addElement(imageCanvasElement)
-
-        // Reset the selected image.
-        selectedImage = nil
     }
 
-    func addPdfCanvasElement() {
-        guard let file = selectedFile else {
-            return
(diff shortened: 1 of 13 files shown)


Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
```diff
@@ -6,6 +6,18 @@ class CanvasViewModel: ObservableObject {
     @Published var annotationCanvas = AnnotationCanvas()
     @Published var selectedImage: UIImage?
 
+    // Selected media
+    @Published var selectedImage: UIImage? {
```
sebastiantoh: There's a duplicate declaration in line 7. Is this intentional?

sebastiantoh (commented):
Haven't tested this PR yet, but from a very preliminary review, it lgtm.

(I'll probably only test this tonight, but if some else has tested this PR, then feel free to merge!)

Regarding some of the decision choices made, I was wondering if it's possible for the `ActiveSheet` enum to also contain an `onDismiss` handler? 
I'm not sure if such an implementation would be susceptible to the "Delay in document picker setting the URL of the picked file" issue.

The reason why I think this could be better (if it works) is that we are guaranteed that onDismiss will only be called once (unless there's a bug with how SwiftUI implements it), whereas `didSet` will probably be triggered more than once (though this isn't really that big a deal)

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
ianyong: Thanks for spotting this. It creeped in after merging `master` into the branch.

ianyong:
> Regarding some of the decision choices made, I was wondering if it's possible for the `ActiveSheet` enum to also contain an `onDismiss` handler?
> I'm not sure if such an implementation would be susceptible to the "Delay in document picker setting the URL of the picked file" issue.

As long as we call the functions to add the canvas elements upon dismissing the sheet, the issue with the delay should be present. Also, I don't think it's possible to do anything with the `ActiveSheet` enum upon dismissing the sheet because the `activeSheet` var would have been set to `nil` already by then.

> The reason why I think this could be better (if it works) is that we are guaranteed that onDismiss will only be called once (unless there's a bug with how SwiftUI implements it), whereas `didSet` will probably be triggered more than once (though this isn't really that big a deal)

Currently, the functions to add the canvas elements already have checks in place such that they should only add canvas elements once. However, let me explore the possibility of passing a callback function into the image picker and document picker directly.

**Update:** The callback functions work great! Thanks for making me think through this more thoroughly. (Brain wasn't working at 2AM :( )

hanstirtaputra (commented):
Tested on my iPad, aside from the mentioned issue of the PDF following the dimensions of the last added canvas element, it was also found that whilst zooming, the pdf would "stutter"?

https://user-images.githubusercontent.com/24553546/111859500-6518b480-897c-11eb-867b-e8123a36a51c.MP4

ianyong:
> Tested on my iPad, aside from the mentioned issue of the PDF following the dimensions of the last added canvas element, it was also found that whilst zooming, the pdf would "stutter"?
> 
>  RPReplay_Final1616216172.MP4

Yea, this is another issue which surfaced after merging #12. Haven't narrowed down what's causing it but I suspect any changes we make to the `PKCanvasView` might also affect this behaviour. As such, I would like to get the repositioning, resizing, etc of canvas elements up first before we relook this issue.

hanstirtaputra:
> > Tested on my iPad, aside from the mentioned issue of the PDF following the dimensions of the last added canvas element, it was also found that whilst zooming, the pdf would "stutter"?
> > RPReplay_Final1616216172.MP4
> 
> Yea, this is another issue which surfaced after merging #12. Haven't narrowed down what's causing it but I suspect any changes we make to the `PKCanvasView` might also affect this behaviour. As such, I would like to get the repositioning, resizing, etc of canvas elements up first.

Alright sounds good! Will merge this in first then.

hanstirtaputra (approved):
LGTM

PR #16, opened 2021-03-20
Fix re-rendering of canvas elements on zoom

ianyong:
Canvas elements no longer re-render unnecessarily.


sebastiantoh (approved):
LGTM. Tested and works well

PR #17, opened 2021-03-20
Create and delete canvas

hanstirtaputra:
## Changes Made
1. Add new canvas
    - This is only a UI implementation, the canvas that is navigated to is not actually added to the `canvases` state. This is to be done after the naming of canvases. Similar to Notability, where new notes are only saved if it is changed (with annotations or note elements) or renamed.
2. Delete existing canvases
    - This allows multi-select delete from `canvases` state using an Edit mode in the `CanvasSelectionView`

## Note
- The state of `canvases` is moved from `HomeView` to `CanvasSelectionView`. This is to reduce the number of callback functions needed to mutate the state of `canvases` from the `CanvasSelectionView`.

image

### DIFF
--- a/Dynavity/Dynavity/view/CanvasSelectionView.swift
+++ b/Dynavity/Dynavity/view/CanvasSelectionView.swift
@@ -78,7 +78,7 @@ struct CanvasSelectionView: View {
                 }) {
                     Image(systemName: "trash")
                 }.disabled(noCanvasSelected)
-                
+
                 SearchBar(text: $searchQuery)
             } else {
                 Button("Edit") {
--- a/Dynavity/Dynavity/view/CanvasSelectionView.swift
+++ b/Dynavity/Dynavity/view/CanvasSelectionView.swift
@@ -26,15 +26,16 @@ struct CanvasSelectionView: View {
         }
     }
 
+    var noCanvasSelected: Bool {
+        canvases.allSatisfy {
+            !$0.isSelected
+        }
+    }
+
     var body: some View {
         NavigationView {
             VStack {
-                HStack {
-                    Button(isEditing ? "Done" : "Edit") {
-                        toggleEditMode()
-                    }
-                    SearchBar(text: $searchQuery)
-                }
+                editButtonGroup
                 .padding()
                 ScrollView {
                     canvasesGrid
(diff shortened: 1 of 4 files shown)


Dynavity/Dynavity/view/canvas/CanvasThumbnailView.swift
```diff
@@ -2,20 +2,36 @@ import SwiftUI
...
-                .resizable()
-                .aspectRatio(contentMode: .fill)
-                .frame(width: 70)
+            if isSelected {
+                // TODO: replace this with canvas thumbnail
+                Image(systemName: "doc")
+                    .resizable()
+                    .aspectRatio(contentMode: .fill)
+                    .frame(width: 70)
+                    .overlay(checkSymbol)
```
sebastiantoh: Does writing something like `.overlay(isSelected ? checkSymbol : nil)` instead work? To remove the code duplication

Dynavity/Dynavity/view/CanvasSelectionView.swift
```diff
@@ -1,8 +1,14 @@
 import SwiftUI
 
 struct CanvasSelectionView: View {
-    @Binding var canvases: [String]
+    struct CanvasDetail: Hashable {
```
sebastiantoh, line 5: Just checking, this is meant to be a temporary substitute for the actual canvas model right? If so, the canvas model wouldn't store `isSelected` I think. Is it possible to save `isSelected` as a state of `CanvasThumbnailView` instead? I think we can just add a `TODO` note for now and we can implement this together with the actual canvas model some other time

sebastiantoh (changes requested):
lgtm, just some minor possible changes

Dynavity/Dynavity/view/canvas/CanvasThumbnailView.swift
hanstirtaputra: Yep that works! Will change it, thanks:)

Dynavity/Dynavity/view/CanvasSelectionView.swift
hanstirtaputra: Yep! I'll add the `TODO` note. As for having `isSelected` as the state of `CanvasThumbnailView`, that would mean that the `CanvasSelectionView` would have to store an array of `CanvasThumbnailView`s instead? I don't think that'll work out.

I think you're right that the `isSelected` shouldn't be in the canvas model. As for that I'll think another way of getting the same behavior once we implement this with the actual canvas model!

sebastiantoh (approved):

PR #19, opened 2021-03-20
Reposition, resize, rotate canvas elements

ianyong:

PR #20, opened 2021-03-20
Add code snippet notes

wltan:
## Features
- Implement the code snippet note type
- Connect to a backend for remote code execution

## Backend Notes
- A backend has been set up to handle code execution remotely. It is adapted from the [Runnable](https://github.com/team-stack-underflow/runnable-backend/wiki) project.
  - Remote code execution is preferred to local code execution as it provides us full control over the execution environment, as opposed to trying to deal with low-level APIs and syscalls on the local system (which may introduce numerous flaws and limitations due to security issues).
- The backend consists of a single WebSocket API. WebSocket is preferred over REST due to its ability to send messages asynchronously.
  - In a REST API, the backend is only allowed to respond [once per request](https://stackoverflow.com/questions/46747854/rest-api-single-request-multiple-responses), thus either the full program output must be available before responding, or an incomplete response will be given. This makes it impractical for programs that block for any reason (e.g. user input, infinite loop, waiting on some resource).
  - In a WebSocket API, once a connection is established both frontend and backend are free to communicate asynchronously with each other. In this case, the frontend sends a launch request to the backend, and the backend sends lines of standard output back to the frontend. This allows the frontend to receive the output line by line, making the program appear more responsive.

## Communication Notes
- The backend ideally accepts requests in the form of JSON objects. The format can be found in the [Runnable API specification](https://github.com/team-stack-underflow/lambdas/blob/master/API%20specification.md).
- However, the frontend and backend currently communicate using string messages instead.
  - In Swift, JSON data is [serialized into binary `Data`](https://github.com/daltoniam/Starscream/issues/283) to be sent through a WebSocket.
  - However, AWS API Gateway [does not handle binary objects well](https://docs.aws.amazon.com/apigateway/latest/developerguide/websocket-api-develop-binary-media-types.html) for WebSocket APIs. There is little documentation on this exact problem as most articles discuss REST APIs instead.
  - Thus, the only remaining solution that works across both systems is to flatten JSON objects into a string to be sent instead. When a response is received it is also as a text message, which needs to be reconstructed into a JSON object.
- The WebSocket client is implemented using the built-in [`URLSessionWebSocketTask`](https://developer.apple.com/documentation/foundation/urlsessionwebsockettask) class. A brief guide on how it may be used can be found [here](https://appspector.com/blog/websockets-in-ios-using-urlsessionwebsockettask).
  - As noted in the above tutorial, it has a very peculiar method of receiving data which requires an explicit infinite recursive loop in order to persist indefinitely.
  - It also has a bad habit of failing (closing the connection) either silently or with cryptic error messages.
- When a launch command is sent, the backend will reply with a `containerId`. Each WebSocket connection can theoretically support multiple concurrent programs, and `containerId` will need to be used to differentiate which output belongs to which program.

## View Notes
- The code snippet model, view, and viewmodel is closely based on the markup text note.
- The right panel is initially at minimum width, as there is no output to be displayed. Once the program has been run and output appears, the right panel's width will scale to fit the output text.
- The text editor uses a monospace font to resemble code.
- The text editor has auto capitalization and autocorrect disabled. This prevents valid code from being autocorrected into something syntactically incorrect.
- Below the text editor are controls to change the execution language and run the program.
- The iPad keyboard produces quotes such as `\u{201c}` (`“`), which is not valid to a number of compilers and interpreters. Thus, when a program is run, each invalid character is replaced with its valid counterpart (such as `"`).

## Known Issues
- In rare cases, the program output may appear out-of-order. Although efforts are made to synchronize the messages from the backend, we cannot control how messages travel from the backend to the frontend, and responses may end up in some strange order. The solution might be to augment each message with a sequential identifier, which requires more work on the backend.
  - This case has not been observed in Dynavity yet, but it has occasionally appeared in Runnable before.

## To Do
- [Done] Ensure that view changes (e.g. code input, set language) update the model.
- [Done] Move the quotation replacement logic such that it can directly update the model string. This will allow changes to be reflected in the view, and minimize problems in future features such as exporting the code to an external location.
- [Done] Implement language-specific templates. C and Java require a strict syntax to be compilable (in particular, Java programs require that the main class be of the arbitrary name `Program`)


hanstirtaputra:
Looks good! I was however unable to test it on my end, I could not seem to focus on the `CodeSnippetView` in order to type and run the code? I faced the same issue with `MarkdownText`.

https://user-images.githubusercontent.com/24553546/111893493-0ca9eb80-8a3e-11eb-9b56-417a0278ad10.mp4

EDIT: 
I think this issue would be fixed with #19

wltan:
> Looks good! I was however unable to test it on my end, I could not seem to focus on the `CodeSnippetView` in order to type and run the code? I faced the same issue with `MarkdownText`.
> 
>  video_2021-03-21_12-07-33.mp4 
> EDIT:
> I think this issue would be fixed with #19

It still works on my end (with a much older base branch). I'll refrain from rebasing until #19 is merged, then rebase and test again.

ianyong (approved):
LGTM!

PR #21, opened 2021-03-21
Fix text persistence issue

sebastiantoh:
**This branch is based off @ianyong 's `ian/reposition-resize-rotate-delete` as I couldn't find a way to disable the annotation gestures, which is necessary for inputting text for testing**
 

# Changes
- Fix text persistence by changing `ObservedObject` to `StateObject` in the canvas element views (this is done to other element views as well, even those that do not have the text persistence issue, for correctness sake)

> A common mistake with @ObservedObjects is to declare and initialise them inside the View itself. This will lead to problems, since every time the @ObservedObject emits an update (one of its @Published properties gets updated), the view will be recreated - which will also create a new @ObservedObject, since it was initialised in the View itself. To avoid this problem, whenever you use @ObservedObject, you always have to inject it into the view. The iOS 14 @StateObject solves this issue.
> - https://stackoverflow.com/questions/61361788/state-vs-observableobject-which-and-when

- Set `@State` and `@StateObject` to private (except for those whose initialisers require knowing of these states)

Spent way too long on this by thinking that the bug was with the `ForEach` :(

@wltan You may want to try incorporating this fix into your PR as well

### DIFF
--- a/Dynavity/Dynavity/view/CanvasSelectionView.swift
+++ b/Dynavity/Dynavity/view/CanvasSelectionView.swift
@@ -8,8 +8,8 @@ struct CanvasSelectionView: View {
     }
 
     @State var canvases: [CanvasDetail]
-    @State var searchQuery: String = ""
-    @State var isEditing = false
+    @State private var searchQuery: String = ""
+    @State private var isEditing = false
 
     let columns = [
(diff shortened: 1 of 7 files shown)


ianyong (approved):
LGTM!

PR #24, opened 2021-03-21
Increase minimum size of code snippets

wltan:
A smaller size causes some elements (e.g. language selection) to appear outside the boundaries of the canvas.

Other changes:
- Properties `minimumWidth` and `minimumHeight` are added to the `CanvasElementProtocol` itself. Otherwise, overriding will not take place and the extension value will always be used.
- Refactored a series of guard statements to make it more compact.


ianyong (approved):
LGTM!

PR #26, opened 2021-03-22
Refactor MarkupElementViewModel

sebastiantoh:
This refactor makes the `MarkupElementViewModel` less coupled with the current HTML conversion logic. With this refactor, the HTML conversion logic will be more substitutable with another method if need be.

## Notes
- The protocol definition in `HtmlExporter` isn't really very ideal since I think it should directly return a `String` representing the exported HTML. However, this doesn't work with an asychronous HTML exporting method (which is being done currently). As such, I've decided to make the protocol method one that requires an asynchronous implementation because:
  1. The current method in `MarkupToHtmlExporter`  is asynchronous
  2. I think it should be relatively trivial for a synchronous implementation to wrap the output in an `AnyPublisher` to conform to the protocol.
- I've also removed `HtmlRenderable` since didn't seem like it was adding much value
- Also, extracted out some String constants into a variable though I didn't do it for everything. Strings like `POST`, `Content-Type` are left as it is.

P.S. Does anyone have any suggestions on where I should place `HtmlExporter` and `MarkupToHtmlExporter`? Doesn't really feel like it belongs entirely in `view-model`


Dynavity/Dynavity/view-model/HtmlExporter.swift
```diff
@@ -0,0 +1,5 @@
+import Combine
+
+protocol HtmlExporter {
```
wltan: The current protocol name doesn't tell us that it is used to render markup text. Perhaps this should be named `MarkupToHtmlExporter` instead?

Then the existing `MarkupToHtmlExporter` could be renamed to specify the implementation used (pandoc).

sebastiantoh: Good point. thanks!

hanstirtaputra (dismissed):
LGTM!

> P.S. Does anyone have any suggestions on where I should place `HtmlExporter` and `MarkupToHtmlExporter`? Doesn't really feel like it belongs entirely in `view-model`

For this perhaps we could have another folder for our networking stuff, eg. `networking` folder? Maybe this can go together with whatever we need for collaboration in the future as well.

sebastiantoh:
> LGTM!
> 
> > P.S. Does anyone have any suggestions on where I should place `HtmlExporter` and `MarkupToHtmlExporter`? Doesn't really feel like it belongs entirely in `view-model`
> 
> For this perhaps we could have another folder for our networking stuff, eg. `networking` folder? Maybe this can go together with whatever we need for collaboration in the future as well.

Sounds good, thanks!!

hanstirtaputra (approved):
LGTM!

PR #27, opened 2021-03-23
Refactor text-related elements

sebastiantoh:
- TextElement is renamed to PlainTextElement
- Text-related elements are no longer composed of a TextElement
- Introduce a new TextElementProtocol, which PlainTextElement,
MarkupElement, CodeElement conform to.


hanstirtaputra (approved):
LGTM!

PR #28, opened 2021-03-23
GraphADT

sebastiantoh:
- Add Graph ADT with the following specifications:
```
 `Graph` is able to represent the following graph types with
 corresponding constraints:
 - Undirected graph
   + An undirected edge is represented by 2 directed edges
 - Directed graph
 - Simple graph
 - Unweighted graph
   + Edges' weights are to set to 1.0
 - Weighted graph

 `Graph` is not able to represent the following graph types:
 - Multigraph

For every Graph g:
 - g is either directed or undirected
 - All nodes in g must have unique labels
 - No self loops present

 Similar to `Node` and `Edge`, `Graph` is a generic type with a type parameter
 `T` that defines the type of the nodes' labels.
```
  - Copy-pasted from my PS1 with minor modifications (BFS, DFS removed, no multigraph support, no negative weights restrictions)
- Copy-pasted corresponding unit tests as well
- Update XcodeGen's `project.yml` to support running tests


hanstirtaputra (approved):
LGTM!

PR #29, opened 2021-03-23
Add annotation palette

hanstirtaputra:
## Changes made:
1. Add annotation palette which will replace PencilKit's `ToolPicker`
    - This solves two things: 
        - Removing unwanted `ToolPicker` tools (ruler and undo)
        - Allows us to continue annotations even after adding PDF/Code/Markdown/LaTeX

## Notes
- The issue with the `ToolPicker` disappearing bug was due to the `PKCanvasWrapperView` no longer being the firstResponder when we add PDF or bring up the keyboard. (The `ToolPicker` will appear when `PKCanvasWrapperView` is the firstResponder)
- Currently, the highlighter and pen tool only supports 4 pen/highlighter widths and 4 colours, we can add more later if time permits

## Known issues
- Annotations always appear behind `CanvasElement`s

IMG_09BF23F2E96E-1


Dynavity/Dynavity/model/AnnotationPalette.swift
```diff
@@ -0,0 +1,43 @@
+import PencilKit
+
+struct AnnotationPalette {
+    static let annotationWidths: [CGFloat] = [3, 6, 9, 12]
+    static let annotationColors = [UIColor.blue, UIColor.red, UIColor.green, UIColor.gray]
+    private let defaultTool = PKInkingTool(.pen, color: .blue, width: 10)
```
sebastiantoh: Should the default width be one of [3, 6, 9, 12]?

Dynavity/Dynavity/model/AnnotationPalette.swift
```diff
@@ -0,0 +1,43 @@
...
+    mutating func switchTool(_ newTool: PKTool) {
+        selectedTool = newTool
+    }
+
+    mutating func switchAnnotationWidth(_ newWidth: CGFloat) {
+        guard let inkingTool = selectedTool as? PKInkingTool else {
+            return
+        }
+
+        // Only pen and and marker (highlighter) inking tools used
```
sebastiantoh, line 24: Can we use a case statement here, and throw a fatalError in the default case? Same for switching colours.  (p.s. i just really like switch statements for enums)

Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
```diff
@@ -4,7 +4,7 @@ import PencilKit
 struct AnnotationCanvasView: View {
     @ObservedObject var viewModel: CanvasViewModel
     @State private var annotationCanvasView = PKCanvasWrapperView()
-    @State private var toolPicker = PKToolPicker()
+    // @State private var toolPicker = PKToolPicker()
```
sebastiantoh: Is this still needed?

Dynavity/Dynavity/view/ToolbarView.swift
```diff
@@ -84,8 +105,112 @@ struct ToolbarView: View {
         }
     }
 
+    private var annotationMenu: some View {
+        HStack {
+            LazyVGrid(columns: columns, spacing: 20) {
+                ForEach(AnnotationPalette.annotationWidths, id: \.self) { width in
```
sebastiantoh: I'm not sure if this counts as interacting directly with the model. Maybe we can shift the annotationWidths to the viewModel? 

Same for colours

Dynavity/Dynavity/view/ToolbarView.swift
```diff
@@ -84,8 +105,112 @@ struct ToolbarView: View {
...
+        .frame(width: 200, height: 100, alignment: .center)
+    }
+
+    private var penSelectionButton: some View {
+        Button(action: {
+            if currentlySelectedTool == .pen {
+                shouldShowAnnotationMenu = true
+            }
+            currentlySelectedTool = .pen
+            viewModel.switchAnnotationTool(PKInkingTool(.pen, color: .blue, width: 10))
```
sebastiantoh: Is it possible to define a default PKInkingTool for pen and marker in viewModel, rather than creating one here?

Dynavity/Dynavity/view/ToolbarView.swift
```diff
@@ -11,12 +12,32 @@ struct ToolbarView: View {
         }
     }
 
+    private enum SelectedAnnotationTool: Identifiable {
+        case pen
+        case highlighter
```
sebastiantoh: i'm not sure if we should use marker here instead? So there's less confusion in terminology

Dynavity/Dynavity/view/ToolbarView.swift
hanstirtaputra: Not sure if the annotationWidths and colors should belong in the viewModel though, feels like it should be part of the palette instead? What do you think of using a getter from the viewModel instead?

Dynavity/Dynavity/view/ToolbarView.swift
hanstirtaputra: Right that sounds good!

Dynavity/Dynavity/view/ToolbarView.swift
sebastiantoh: > What do you think of using a getter from the viewModel instead?

Yep, I think this sounds great!

Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
hanstirtaputra: Nope, I missed this one, thanks for pointing it out!

Dynavity/Dynavity/view/ToolbarView.swift
hanstirtaputra: What do you think of defining it in AnnotationPalette in the model, and using a getter from the viewModel? I feel like the default definitions don't really belong in the viewModel?

sebastiantoh: Yep I think it sounds good to leave the definitions in the model. But maybe instead of a getter, the viewModel can have an overloaded method / another method that selects the default from the model?

hanstirtaputra: Alright! Sounds good. Not too sure if you meant a generic method that selects default from any model, or just for the annotationTool. Nonetheless I've made some updates! Will change again if necessary:)

sebastiantoh (approved):
LGTM!

PR #30, opened 2021-03-24
Add factory pattern

hanstirtaputra:
## Changes made:
1. Refactor `CanvasElementMapView` to apply factory pattern in view creation
2. Remove direct access of model from `CanvasElementMapView` by using a getter through `CanvasViewModel`

### DIFF
--- a/Dynavity/Dynavity/view/CanvasElementMapView.swift
+++ b/Dynavity/Dynavity/view/CanvasElementMapView.swift
@@ -1,14 +1,8 @@
-//
-//  CanvasElementMapView.swift
-//  Dynavity
-//
-//  Created by Hans Sebastian Tirtaputra on 15/3/21.
-//
-
 import SwiftUI
 
 struct CanvasElementMapView: View {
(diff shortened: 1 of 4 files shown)


Dynavity/Dynavity/view/CanvasElementMapView.swift
```diff
@@ -18,22 +12,7 @@ struct CanvasElementMapView: View {
     }
 
     private func transformToView(element: CanvasElementProtocol) -> some View {
```
sebastiantoh: Do you think we should remove this function and directly use the view factory in the foreach?

sebastiantoh (dismissed):
LGTM!

Dynavity/Dynavity/view/CanvasElementMapView.swift
hanstirtaputra: Sounds good! Thanks

sebastiantoh (approved):
lgtm

PR #31, opened 2021-03-25
Add mode for interacting with canvas elements

ianyong:
Also made the following changes:
* Moved button handlers out of `CanvasView`, into `CanvasViewModel` to better conform to the MVVM architecture.
* Fixed the annotation menu not closing upon switching tools.
  * Previously, if you opened the annotation menu on either the pen/marker tool, then switch to the other tool, the annotation menu would appear for the other tool as well.

### DIFF
--- a/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
+++ b/Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
@@ -11,6 +11,7 @@ class CanvasViewModel: ObservableObject {
     @Published var scaleFactor: CGFloat = 1.0
     @Published var selectedCanvasElementId: UUID?
     @Published var currentlySelectedTool: ToolbarView.SelectedAnnotationTool?
+    @Published var shouldShowAnnotationMenu = false
 
     // Reposition drag gesture
     private var dragStartLocation: CGPoint?
@@ -192,15 +193,15 @@ extension CanvasViewModel {
         currentlySelectedTool != nil
     }
 
-    func switchAnnotationTool(_ newTool: PKTool) {
+    private func switchAnnotationTool(_ newTool: PKTool) {
         annotationPalette.switchTool(newTool)
     }
 
-    func switchAnnotationWidth(_ newWidth: CGFloat) {
+    private func switchAnnotationWidth(_ newWidth: CGFloat) {
         annotationPalette.switchAnnotationWidth(newWidth)
     }
 
-    func switchAnnotationColor(_ newColor: UIColor) {
+    private func switchAnnotationColor(_ newColor: UIColor) {
         annotationPalette.switchAnnotationColor(newColor)
     }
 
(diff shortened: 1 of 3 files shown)


Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
```diff
@@ -186,15 +196,19 @@ extension CanvasViewModel {
 
 // MARK: Annotation palette controls
 extension CanvasViewModel {
-    func switchAnnotationTool(_ newTool: PKTool) {
+    var isAnnotationPaletteActive: Bool {
```
sebastiantoh: Would it be better to have a `currentMode` variable instead which stores an enum? I think it makes the code easier to reason about and more future-proof esp if we were to add more modes that are mutually exclusive

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
```diff
@@ -223,3 +237,42 @@ extension CanvasViewModel {
         }
     }
 }
+
+// MARK: Annotation palette controls button handlers
+extension CanvasViewModel {
+    func selectNoAnnotationTool() {
```
sebastiantoh: I think this function could be renamed to something like `clearSelectedAnnotationTool`?

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
ianyong: Does this mean that we will have 2 levels of enums? i.e. one for mode, another for the annotation tools (for the annotation mode)?

sebastiantoh: ~~That was my original intention, but now that I think about it, we could actually merge the two:~~

```swift

enum CanvasMode {
  case selection
  case annotation(ToolbarView.SelectedAnnotationTool.pen)
  // possibly more modes e.g. deletion where you can single tap to delete and stuff
}
```

NVM, I realised we can't merge them. We would still have two enums. But I don't think having 2 enums is that bad. We could probably only need 2 enum definitions, but only 1 enum state.

EDIT: maybe we can have just 1 enum by storing PKTool directly:

```swift
enum CanvasMode {
  case selection
  case annotation(PKTool)
  // possibly more modes e.g. deletion where you can single tap to delete and stuff
}
```

ianyong: Storing the `PKTool` directly isn't possible since we need to distinguish between the various `InkType`s (which are strings) for `PKInkingTool`. Would it make more sense to simply use a flattened enum? i.e. treat each annotation tool as a mode in itself.

For reference: https://developer.apple.com/documentation/pencilkit/pkinkingtool/inktype

sebastiantoh: opps, my bad.

> Would it make more sense to simply use a flattened enum? i.e. treat each annotation tool as a mode in itself.

Yep, I think this makes sense. Also, I realise we probably need to store some state for the lastSelectedPen and lastSelectedHighlighter, unless we don't mind it resetting to the default pen / highlighter everytime.

I'm not sure if it's worth refactoring given our time constraint. I think I'll just approve this PR first and we can refactor this if there's time? Feel free to merge, unless you want to refactor it now

sebastiantoh (dismissed):
lgtm

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
ianyong: > Also, I realise we probably need to store some state for the lastSelectedPen and lastSelectedHighlighter, unless we don't mind it resetting to the default pen / highlighter everytime.

I took a look at this and it seems that the storing of tool state should probably be implemented within `AnnotationPalette` itself. Seems very out of scope for this PR though.

sebastiantoh (approved):
lgtm, thanks!

PR #32, opened 2021-03-26
Add Backlink component

sebastiantoh:
Two distinct components of interest in this PR:
1. Graph library
2. Backlink library

**Backlink --> Graph**
- The Backlink library uses the Graph library internally, with the `BacklinkGraph` protocol acting as the interface between the two. The Backlink library adds an extension to `Graph` to conform to `BacklinkGraph`.

**Backlink library**
- This is designed to be as general as possible, similar to what was done for our physics engine
- It supports adding of backlinks between items of different types. This is done by using a type-erased `NamedIdentifiable`: `AnyNamedIdentifiable`
   - One possible issue I can see with this is difficulty in retrieving the actual type. But type-erasure seems to be the only way  to store a heterogeneous collection of some protocol
- Stuff that are exposed to users of the backlink library:
  - `NamedIdentifiable`: users will need to ensure that their models conform to this protocol before they can add backlink between their models. 
  - `AnyNamedIdentifiable`: users will need to wrap their `NamedIdentifiable` models in an `AnyNamedIdentifiable` when adding using the `BacklinkEngine`
  - `BacklinkEngine`: this is where users add backlinks and retrieve backlinks


ianyong (approved):
LGTM!

PR #33, opened 2021-03-27
Store state per annotation tool

ianyong:
See the original discussion [here](https://github.com/Dynavity/dynavity/pull/31#discussion_r602216522).

### DIFF
--- a/Dynavity/Dynavity/model/AnnotationPalette.swift
+++ b/Dynavity/Dynavity/model/AnnotationPalette.swift
@@ -1,54 +1,61 @@
 import PencilKit
 
 struct AnnotationPalette {
-    static let annotationWidths: [CGFloat] = [3, 6, 9, 12]
-    static let annotationColors = [UIColor.blue, UIColor.red, UIColor.green, UIColor.yellow]
-    static let defaultPenTool = PKInkingTool(.pen, color: .blue, width: annotationWidths[2])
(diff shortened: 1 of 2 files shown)


Dynavity/Dynavity/model/AnnotationPalette.swift
```diff
@@ -1,54 +1,61 @@
...
-        guard let inkingTool = selectedTool as? PKInkingTool else {
-            return
+    mutating func setAnnotationWidth(_ newWidth: CGFloat) {
+        switch selectedTool {
+        case .pen:
+            penTool.width = newWidth
+        case .marker:
+            markerTool.width = newWidth
+        default:
+            // Do nothing.
```
sebastiantoh: I would prefer to `fatalError` here instead since it helps us to verify that end-users are not allowed to change the width when a non-pen / marker tool is selected. And if for some reason end-users are able to do so (e.g. a lapse in our UI), we would be immediately notified rather than things failing silently.

Likewise for setting of annotation colour

Edit: maybe `assertionFailure` would be better in capturing the intent rather than `fatalError`

ianyong: That makes sense, thanks!

sebastiantoh (approved):
LGTM, thanks!!

PR #34, opened 2021-03-28
Fix scroll to top issue

hanstirtaputra:
Fix issue where interacting with the toolbar causing the canvas elements to "disappear". This happens when the `AnnotationCanvas` scrolls to the top when we tap on the status bar.


ianyong (approved):
LGTM! (I didn't even know this was an issue LOL)

PR #35, opened 2021-03-29
Fix image resizing

ianyong:
Was planning on locking the aspect ratio when dragging as well, but it's much more complicated than I thought. Can be added if we have the time in Sprint 3.


sebastiantoh (approved):
LGTM!

PR #36, opened 2021-03-29
Add drag overlay control

ianyong:
Unfortunately, there doesn't seem to be a good way of both consuming and selectively propagating touch events to views underneath the current view in both SwiftUI and UIKit. This fixes the issue of being unable to scroll the contents of selected elements.


sebastiantoh (approved):
LGTM!

PR #37, opened 2021-03-29
Add graph view

sebastiantoh:
Changes:
- Refactor backlink component
  - Previous implementation had a lot of unnecesary "layering" and complexity. 
  - It also had difficulties supporting encoding of the models for persistence storage since the Node and Edge objects of the Graph library will not be exposed to the main Dynavity application. With that, I've created another BacklinkNode and BacklinkEdge class, which will be exposed to the main Dynavity application and can be encoded / decoded.
- Add a graph view (accessible by tapping the honeycomb look-alike button at top right hand corner of the canvas selection screen)
  - Supports panning, zooming and dragging of nodes
  - Supports searching of nodes (note: this only highlights nodes matching the search term, rather than filtering nodes)
  - Don't mind the styling, the dark background is so I know where exactly the canvas ends

P.S. I added some non-operator extensions to `CGPoint+Operators`. I hope that's fine. Also, does anyone know how we can better organise these project-wide extensions?


Dynavity/Dynavity/view-model/graph/GraphMapViewModel.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    func getNodes() -> [BacklinkNode] {
+        backlinkEngine.nodes
+    }
+
+    func getEdges() -> [BacklinkEdge] {
+        backlinkEngine.edges
+    }
+
+    private func initialiseBacklinkEngine() {
```
ianyong, line 24: This is a temporary function for testing, right? Might be good to add a `TODO`.

Dynavity/Dynavity/view-model/graph/GraphMapViewModel.swift
```diff
@@ -0,0 +1,83 @@
...
+        let nodes: [BacklinkNode] = (0..<20).map { i in
+            BacklinkNode(id: UUID(), name: "Node \(i)", position: CGPoint(x: Int.random(in: 0...700),
```
ianyong, line 26: Would it be better to create an initialiser for `BacklinkNode` that initialises its own `id` property?

Dynavity/Dynavity/view/CanvasSelectionView.swift
```diff
@@ -38,9 +43,15 @@ struct CanvasSelectionView: View {
...
-                    canvasesGrid
+                switch selectionMode {
+                case .grid:
+                    ScrollView {
+                        canvasesGrid
+                    }
+                case .graph:
+                    GraphView(searchQuery: $searchQuery)
                 }
+
```
ianyong: Extraneous newline?

Dynavity/Dynavity/view/CanvasSelectionView.swift
```diff
@@ -67,6 +78,24 @@ struct CanvasSelectionView: View {
...
+                Image(systemName: "circle.grid.hex.fill")
+            }
+        case .graph:
+            return Button(action: {
+                selectionMode = .grid
+            }) {
+                Image(systemName: "square.grid.2x2.fill")
+            }
+        }
+
```
ianyong: Here too.

Dynavity/Dynavity/view/graph/GraphView.swift
```diff
@@ -0,0 +1,118 @@
...
+                if viewModel.selectedNode == nil {
+                    self.dragOffset = .zero
+                    self.originOffset += value.translation
+                } else {
+                    viewModel.handleNodeDragEnd(value, viewportZoomScale: zoomScale)
+                }
+            }
+    }
+
+    private var viewportMagnificationGesture: some Gesture {
```
ianyong, line 88: This doesn't seem to zoom according to the location where the screen is pinched.

ianyong (dismissed):
Mostly LGTM!

> P.S. I added some non-operator extensions to `CGPoint+Operators`. I hope that's fine.

I think it's fine since it's still dealing with `CGPoint` operations. Perhaps the name of the file can be changed if you wish.

> Also, does anyone know how we can better organise these project-wide extensions?

Hmm, not too sure whether there is even a good universal way of dealing with these. I guess for extensions of library structs/classes (as is `CGPoint`), we can place them in a root `extensions` folder?

Dynavity/Dynavity/view-model/graph/GraphMapViewModel.swift
sebastiantoh: Yep, will do! I'll leave the initialiser here as it is though, since it'll be replaced eventually

Dynavity/Dynavity/view/CanvasSelectionView.swift
sebastiantoh: Oops, thanks for pointing it out!

Dynavity/Dynavity/view/graph/GraphView.swift
sebastiantoh: Oops, I knew something didn't feel right. Just didn't know what exactly.
Unfortunately, I think this fix isn't something easy since MagnificationGesture doesn't give the location. I'll fix this some other time (hopefully) if that's OK

ianyong (approved):
LGTM! Thanks!

PR #38, opened 2021-03-29
Refactor CodeElementViewModel

wltan:
Similar to #26, but for code elements. This makes `CodeElementViewModel` less coupled with the low-level WebSocket implementation of the code execution, allowing the implementation to be easily substituted in future (e.g. by executing locally instead).

## Notes

- We want execution output to be generated line by line, as opposed to all at once when the execution has terminated. This makes the user program appear more responsive, and will still show some output even if the program does not eventually terminate (e.g. stuck in infinite loop or blocking on user input).
  - Thus, we want an asynchronous publisher that is able to stream the output line by line.
  - The concrete implementation `RunnableCodeExecutor` uses [`PassthroughSubject`](https://developer.apple.com/documentation/combine/passthroughsubject) as the publisher, but it is not known to the rest of the client as it has been erased into an `AnyPublisher`.
- The client (`CodeElementViewModel`) subscribes to the publisher using `sink`, which processes each incoming element (line of output) once available.

### DIFF
--- a/Dynavity/Dynavity/networking/CodeExecutor.swift
+++ b/Dynavity/Dynavity/networking/CodeExecutor.swift
@@ -1,0 +1,5 @@
+import Combine
+
+protocol CodeExecutor {
+    func getOutputPublisher(program: String, language: CodeElement.CodeLanguage) -> AnyPublisher<String, Never>
+}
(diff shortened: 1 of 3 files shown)


sebastiantoh (approved):
LGTM!

PR #39, opened 2021-03-29
Enable scrolling on code execution output

wltan:
If the output of a code execution spans many lines, the bottom is truncated to ... as the output area is not scrollable. This can be fixed by wrapping the output in a `ScrollView`.

## Notes
- The basic change is very simple: https://github.com/Dynavity/dynavity/commit/db98e840242571e467e386b54d7b3b22513505aa
- However, this causes SwiftLint to complain about the length of the closure body, so I had to split the `body` into some smaller components.
  - Despite `outputWindow` being much smaller/simpler compared to `codeWindow`, both have been extracted to their respective components in order to keep a similar level of abstraction within the `body`.


sebastiantoh (approved):
LGTM

PR #40, opened 2021-03-30
Add navigation to canvas from graph

sebastiantoh:
Changes:
- Add ability to navigate to canvas from graph view by long press on the relevant nodes 
- Renamed `selectedNode` to `draggedNode` to prevent confusion


ianyong (approved):
LGTM!

PR #41, opened 2021-03-31
Add TodoElement

ianyong:
Due to conflicting gestures, it was not possible to use SwiftUI's `List`. As such, I had to implement a custom list from the ground up. A known issue is that scrolling the list while swiping diagonally left and downwards causes the delete button to poke out a bit without being reset.


sebastiantoh (approved):
LGTM! Thanks for the variable renaming!

PR #43, opened 2021-03-31
Add uml diagramming

hanstirtaputra:
## Changes made:
1. Add UML Diagramming feature which includes
    -  Side menu to add UML shapes. 
        - Current PR has only two shapes, more will be added in a subsequent PR to prevent this PR from being too large.
    - Connecting two UML shapes with a line. 
       - Current line is undirected
       - Current line cannot be delete once it has been added
       - When repositioning a UMLElement, the line will only be updated after the element has been repositioned, and will not update as the element is being moved. This prevents the repositioning from feeling unresponsive due to slow processing time to update the line. (Will work on making the update more efficient if time permits)
    - UML shape text can be edited
2. Add unit tests for models related to UML digram connection

## Testing this feature:
- Comment/remove `assert(checkRepresentation())` in `Graph.swift`, otherwise the UML connection algorithm will run slowly.

## Notes:
To connect two UML shapes:
1. Tap on UMLElement to select it
2. Tap on one of the four connection points
3. Tap on another connection points on any other UMLElement, this will form a connection between the first element in step 2. and the current tapped element.

## Known issues:
- UML connector lines move faster than UMLElements when the keyboard shows. This is due to the connectors being drawn before the UMLElements
- UML connector lines does not account for rotated shapes

## Design decisions:
1. Modeling UML diagrams:
    - Alternative 1: Using protocols to represent UML shapes as CanvasElements. `CanvasElementProtocol` <-- `UmlElementProtocol` <-- `SomeUmlElementProtocol`
        - This is quite redundant as the different `SomeUmlElementProtocol` will all have the same functionality in the context of Dynavity since we only care about the shape (how its drawn) and the text that is inside the shape. 
    - Alternative 2 (Chosen): Only use one `UmlElementProtocol`, and have different types of UML represented by Enum
         - This allows us to know what UML type it is (EG. class diagram, activity diagram, etc.) to show separate them in the View.
2. Representing UML Connectors and UMLElements in view:
    - Alternative 1: Store `UmlConnector` and `UmlElementProtocol` in a separate layer, for example `CanvasUmlElementMapView`
        - This achieves separation of concerns as a `UmlConnector` is not really a `CanvasElementProtocol` as it does not care about the rotation, height, width, etc. of a `CanvasElementProtocol`.
        - However, `UmlElementProtocol` is very similar to `CanvasElementProtocol`. For example the gestures and the `SelectionOverlayView` is the same. Thus having `UmlElementProtocol` and `CanvasElementProtocol` in two separate layers is unnecessary complexity.
    - Alternative 2 (Chosen): Store `UmlConnector` in `CanvasElementMapView` as well
        - The disadvantage is that `UmlConnector` can arguably be considered as not a `CanvasElement`, and thus should not belong in the `CanvasElementMapView`
        - Another disadvantage is that the `CanvasElementMapView` might get bloated and do too many things
        - Nonetheless, the advantage is that we do not unnecessarily separate `UmlElementProtocol` and `CanvasElementProtocol`.  Since `UmlElementProtocol` is a `CanvasElementProtocol` and they share a lot of the same functionalities.

IMG_8441A8C29E70-1


sebastiantoh (commented):
I think we can move the `graph` directory to be on the same level as `model`, `view-model`, etc. Likewise for the `graph` directory in the Tests directory (P.S. I initially shifted it to be in the `backlink` directory because I didn't think anybody else would be using it)

Dynavity/Dynavity/backlink/graph/Graph.swift
```diff
@@ -52,6 +52,25 @@ struct Graph<T: Hashable> {
...
+    func getNode(_ node: Node<T>) -> Node<T>? {
+        if !containsNode(node) {
+            return nil
+        }
+        return nodes.first { $0 == node }
+    }
+
+    /// Gets edge between source and destination node if it exists, otherwise return empty optional.
+    func getEdgeBetween(source: Node<T>, destination: Node<T>) -> Edge<T>? {
+        for edge in edges {
```
sebastiantoh, line 66: Not sure if this is micro-optimizing, but I think it may be slightly more efficient to iterate from the edges in `adjList[source]` directly

sebastiantoh (dismissed):
Not really sure what's going on, but LGTM!

sebastiantoh (approved):

PR #44, opened 2021-04-02
Set up CocoaPods/Firebase

wltan:
This PR sets up the basic Firebase dependency using CocoaPods, which will (hopefully) be used for the collaboration feature.

Based on my experience so far, there may be some potential issues when working across multiple developers. Thus I'd like to resolve them before further changes are committed. Things to check:
1. Checkout this branch
2. Open the project in Xcode
3. Check if `import Firebase` works in `DynavityApp.swift`
4. If there is a build error, close the project, run `pod install`, and reopen the project.
  a. Do **not** re-run `xcodegen` after `pod install`. Otherwise, you should get the error `No such module 'Firebase'`.
  b. Open the project using `Dynavity.xcworkspace` and **not** `Dynavity.xcodeproj`. Otherwise, you should get the error `Could not build Objective-C module 'Firebase'`.

Once the above behavior can be replicated and verified, I will update the setup instructions in `README.md`

## Notes
- It is recommended that `Podfile`, `Podfile.lock`, and `Pods/` are all checked into version control. This ensures that everyone is using the same set of dependencies.
  - However, if a `pod install` is still required to make things work in Xcode then there is no point checking in `Pods/` and it can be removed.

https://github.com/Dynavity/dynavity/blob/a286c63f5fe1522a1bea4eb030414ebeb3fa036b/.gitignore#L51-L57


ianyong:
CocoaPods installation instruction for ARM architecture (for documentation purposes):
```sh
sudo arch -x86_64 gem install ffi
arch -x86_64 pod install
```

ianyong:
> This PR sets up the basic Firebase dependency using CocoaPods, which will (hopefully) be used for the collaboration feature.
> 
> Based on my experience so far, there may be some potential issues when working across multiple developers. Thus I'd like to resolve them before further changes are committed. Things to check:
> 
> 1. Checkout this branch
> 2. Open the project in Xcode
> 3. Check if `import Firebase` works in `DynavityApp.swift`
> 4. If there is a build error, close the project, run `pod install`, and reopen the project.
>    a. Do **not** re-run `xcodegen` after `pod install`. Otherwise, you should get the error `No such module 'Firebase'`.
>    b. Open the project using `Dynavity.xcworkspace` and **not** `Dynavity.xcodeproj`. Otherwise, you should get the error `Could not build Objective-C module 'Firebase'`.
> 
> Once the above behavior can be replicated and verified, I will update the setup instructions in `README.md`

Can confirm that I ran into all the issues listed above in the exact same order. Following the instructions resolved the issues.

> ## Notes
> * It is recommended that `Podfile`, `Podfile.lock`, and `Pods/` are all checked into version control. This ensures that everyone is using the same set of dependencies.
>   
>   * However, if a `pod install` is still required to make things work in Xcode then there is no point checking in `Pods/` and it can be removed.
> 
> https://github.com/Dynavity/dynavity/blob/a286c63f5fe1522a1bea4eb030414ebeb3fa036b/.gitignore#L51-L57

I had to run `pod install` to get things to work.

Dynavity/Podfile
```diff
@@ -0,0 +1,20 @@
...
+
+  target 'DynavityTests' do
+    inherit! :search_paths
+    # Pods for testing
+  end
+
+  target 'DynavityUITests' do
+    # Pods for testing
+  end
+
```
ianyong:
```suggestion
```

ianyong (commented):
Minor nit.

ianyong (approved):
LGTM!

PR #45, opened 2021-04-02
Update SideMenu view to support display and adding / removal of linked canvases (UI Implementation)

sebastiantoh:
### Changes
- Update SideMenu view to support display and adding / removal of linked canvases (note: this is just a UI implementation)
- Side menu can now be dismissed on tap (by tapping anywhere on the black translucent overlay), rather than just a left swipe
- Update the graph styling
- Shift bunch of extensions into extensions folder
- Set GraphMapViewModel to instantiated in the root view, which is passed to the relevant child views (GraphView and SideMenuView) via environment object. This is in preparation for integration of the actual canvases with the backlink engine
  - P.S. I'm thinking of doing something similar for `CanvasViewModel` so that the individual elements will be able to figure out whether they are selected or not (this is mainly for the markup text elements: when they are not selected, the text editor segment should be hidden). But I'm not sure if this is good practice since this is similar to having global state


wltan (approved):
Thanks for moving the extension code too!

PR #46, opened 2021-04-02
Minor tweaks to WebView

sebastiantoh:
- Set user agent of the WebView to be an iPhone + add HTML template that takes into account viewport so that the display text size is acceptable
- Fix a SwiftUI warning about publishing changes on background thread


hanstirtaputra (approved):
LGTM!

PR #47, opened 2021-04-02
UML feature bug fixes

hanstirtaputra:
## Changes made:
- Update UML connection to account for rotation, resizing, and unselecting elements. Note that the connection formed on a rotated UML element still has some quirks at certain rotation angles (the line might be a straight line instead) 
- Update unit test cases to remove some assertions

I think that I will work on a better solution (such as updating the connection anchor point as the UML element rotates) after working on other more important features (eg. deleting of connection lines).


ianyong (commented):
Getting a huge delay between rotating/moving UML elements and the UML connections forming. Is this intended?

hanstirtaputra:
> Getting a huge delay between rotating/moving UML elements and the UML connections forming. Is this intended?

We need to comment out `assert(checkRepresentation())` in `Graph.swift` to avoid this delay!

ianyong:
> > Getting a huge delay between rotating/moving UML elements and the UML connections forming. Is this intended?
> 
> We need to comment out `assert(checkRepresentation())` in `Graph.swift` to avoid this delay!

Ah ok, thanks!

ianyong (approved):
LGTM! Just some quirks with the lines at certain angles as mentioned in the PR description.

simulator_screenshot_52881599-5BCA-4DDA-AD77-894C0BD4C04A

PR #48, opened 2021-04-02
Delete canvas elements

ianyong:
I tried to use context menu, but it is extremely buggy. See this [StackOverflow](https://stackoverflow.com/questions/65118408/swiftui-contextmenu-flipping-bug) discussion for more information. You can also checkout the branch `ian/add-context-menu-delete` to see for yourself. Not only does the context menu not respect rotation, it also doesn't render the contents of more complicated elements such as PDFs and code blocks. After spending quite some time trying to make this work, I don't think it is possible (or at least not without introducing a lot of hacks).

I ended up adding the delete button to `SelectionOverlayView` as yet another extended control. Upon tapping on it, the user is prompted with an alert to confirm their action.

### DIFF
--- a/Dynavity/Dynavity/model/canvas/Canvas.swift
+++ b/Dynavity/Dynavity/model/canvas/Canvas.swift
@@ -18,6 +18,14 @@ struct Canvas {
         canvasElements.append(element)
     }
 
+    mutating func removeElement(_ element: CanvasElementProtocol) {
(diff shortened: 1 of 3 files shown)


hanstirtaputra (approved):
LGTM!

Just a suggestion, perhaps we could remove some of the resize controls in the future if the amount of controls feels overwhelming. Maybe something like this:

IMG_09A0F704A730-1

PR #49, opened 2021-04-02
Add new canvas elements to center of viewport

ianyong:
Did some refactoring as well to make the code cleaner. The math behind this is a bit complicated...


hanstirtaputra (approved):
Yep the math does look complicated... LGTM!

PR #50, opened 2021-04-03
Add panning & zooming in canvas mode

ianyong:
Also added the ability to clear the canvas element selection by tapping anywhere on the canvas.

This was achieved by adding a new mode to `AnnotationCanvasView` in which its active `PKTool` is set to the eraser. I tried a few other methods, such as removing the `drawingGestureRecognizer` but those didn't seem to work. Once again, there is a memory issue that occurs only when connected to the debugger and panning & zooming the canvas very quickly.

### DIFF
--- a/Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
+++ b/Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
@@ -65,8 +65,13 @@ extension AnnotationCanvasView: UIViewRepresentable {
     }
 
     func updateUIView(_ uiView: PKCanvasView, context: Context) {
-        annotationCanvasView.zoomScale = viewModel.scaleFactor
-        annotationCanvasView.contentOffset = viewModel.canvasTopLeftOffset
+        if annotationCanvasView.zoomScale != viewModel.scaleFactor {
+            annotationCanvasView.zoomScale = viewModel.scaleFactor
+        }
+
+        if annotationCanvasView.contentOffset != viewModel.canvasTopLeftOffset {
+            annotationCanvasView.contentOffset = viewModel.canvasTopLeftOffset
+        }
 
         // Disable drawing by setting the tool to the eraser.
         if isDrawingDisabled {
--- a/Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
+++ b/Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
@@ -4,9 +4,15 @@ import PencilKit
 struct AnnotationCanvasView: View {
     @ObservedObject var viewModel: CanvasViewModel
     @State private var annotationCanvasView = PKCanvasWrapperView()
+    private let isDrawingDisabled: Bool
+
+    init(viewModel: CanvasViewModel, isDrawingDisabled: Bool) {
+        self.viewModel = viewModel
+        self.isDrawingDisabled = isDrawingDisabled
+    }
 
     init(viewModel: CanvasViewModel) {
-        self.viewModel = viewModel
+        self.init(viewModel: viewModel, isDrawingDisabled: false)
     }
 }
 
@@ -59,6 +65,12 @@ extension AnnotationCanvasView: UIViewRepresentable {
     }
 
     func updateUIView(_ uiView: PKCanvasView, context: Context) {
+        // Disable drawing by setting the tool to the eraser.
+        if isDrawingDisabled {
+            annotationCanvasView.tool = PKEraserTool(.vector)
+            return
+        }
+
         // Update annotation tool
         annotationCanvasView.tool = viewModel.getCurrentTool()
     }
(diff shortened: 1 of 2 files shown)


Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
```diff
@@ -59,6 +65,20 @@ extension AnnotationCanvasView: UIViewRepresentable {
     }
 
     func updateUIView(_ uiView: PKCanvasView, context: Context) {
+        if annotationCanvasView.zoomScale != viewModel.scaleFactor {
+            annotationCanvasView.zoomScale = viewModel.scaleFactor
+        }
+
+        if annotationCanvasView.contentOffset != viewModel.canvasTopLeftOffset {
+            annotationCanvasView.contentOffset = viewModel.canvasTopLeftOffset
+        }
```
sebastiantoh: What are these lines for? Adding them in seems to make the panning and zooming quite buggy

ianyong: This is so that we can update the viewport of the `AnnotationCanvasView` when some other view pans/zooms the canvas.

Maybe we can leave this PR open for now as I agree that the panning/zooming isn't really up to standard. I'll try to figure out some workaround for this issue.

sebastiantoh: > This is so that we can update the viewport of the AnnotationCanvasView when some other view pans/zooms the canvas.

I see. But right now, there are no other views that pans / zooms the canvas right? Are there plans to add them?

ianyong: > I see. But right now, there are no other views that pans / zooms the canvas right? Are there plans to add them?

Right now, it's for panning and zooming when in the canvas element mode. Unfortunately, it isn't possible to use the same `AnnotationCanvasView` for this as well (through manipulating the `zIndex`) because then the annotations would go underneath the canvas elements when in the canvas element mode. (Unless we brand this as a feature? LOL)

If we do go with the above though, then it would solve this issue.

sebastiantoh: Wait, by panning and zooming in canvas element mode, you mean using 2 fingers right? Similar to when an annotation tool is selected? Right now, if I comment out these lines, the panning and zooming still works in canvas element mode?

ianyong: > Wait, by panning and zooming in canvas element mode, you mean using 2 fingers right? Similar to when an annotation tool is selected? Right now, if I comment out these lines, the panning and zooming still works in canvas element mode?

Yup! If you comment out those lines, then your annotations won't move though (and your canvas elements will snap back once you switch to an annotation mode).

sebastiantoh: Oh, I see. I didn't realise you created 2 annotation canvas views. If only there was an actual way to disable annotations on the PKCanvas...

ianyong: > Oh, I see. I didn't realise you created 2 annotation canvas views. If only there was an actual way to disable annotations on the PKCanvas...

I think I found a way that addresses this issue! It still suffers from the memory issue when the debugger is active however.

sebastiantoh: Wait, actually can you try using the same annotation canvas view, but setting the annotation tool to be some transparent pen? Then whenever we store the drawing, we can filter the strokes that are transparent.

I did some preliminary testing and the filtering based on strokes does work

ianyong: > Wait, actually can you try using the same annotation canvas view, but setting the annotation tool to be some transparent pen? Then whenever we store the drawing, we can filter the strokes that are transparent (or not, since it doesn't really affect user experience)

The issue with this is that we wouldn't be able to interact with the canvas elements since the `AnnotationCanvasView` would consume all the gestures. Alternatively, if we put the `AnnotationCanvasView` behind the `CanvasElementMapView`, then the annotations would not appear over the canvas elements.

sebastiantoh: > I think I found a way that addresses this issue! It still suffers from the memory issue when the debugger is active however.

Great! Let me know if you want to try what I suggested above ^. If not, I think the current implementation LGTM!

sebastiantoh: > The issue with this is that we wouldn't be able to interact with the canvas elements since the AnnotationCanvasView would consume all the gestures. Alternatively, if we put the AnnotationCanvasView behind the CanvasElementMapView, then the annotations would not appear over the canvas elements.

Opps, you're right!

sebastiantoh (approved):
LGTM!

PR #51, opened 2021-04-03
UML connector deletion

hanstirtaputra:
## Changes made:
- Refactor `CanvasViewModel` to extract out functions related to UML diagramming feature
- Add ability to delete UML connectors after they are formed

## Notes:
To delete a UML connector:
1. Tap on the UML connector (Tap on selected UML connector again to unselect)
2. The selected connector will be highlighted, and a trash icon will appear
3. Tap on the trash icon to delete the connector.

Unlike deletion of `CanvasElementProtocol`s, the deletion of `UmlConnector`s does not require the user to confirm the deletion from a pop-up alert. I think it is not necessary for `UmlConnector` as they can easily be formed again! I also feel that accidental deletions are unlikely to occur.

Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2021-04-03 at 18 35 55

### DIFF
--- a/Dynavity/Dynavity/model/canvas/Canvas.swift
+++ b/Dynavity/Dynavity/model/canvas/Canvas.swift
@@ -76,4 +76,12 @@ extension Canvas {
 
         umlConnectors[index] = connector
     }
+
+    mutating func removeUmlConnector(_ connector: UmlConnector) {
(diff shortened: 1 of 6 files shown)


sebastiantoh (approved):
LGTM!

PR #54, opened 2021-04-04
Fix choppiness with collaboration when dragging/resizing/rotating

ianyong:
See https://github.com/Dynavity/dynavity/pull/53#issuecomment-812982458 for more context on this issue.

When updating from Firebase, we exempt the currently selected canvas element from being updated. This will potentially result in the state not being exactly the same across multiple devices but I think it isn't a problem as future updates when the canvas element is not selected will result in it being updated.


sebastiantoh (approved):
LGTM, thanks!

PR #55, opened 2021-04-04
Enlarge hitbox of uml connectors

hanstirtaputra:
This fixes the issue of user being unable to select UML connectors


ianyong (approved):
LGTM!

PR #56, opened 2021-04-04
Implement element-level synchronization

wltan:
Observer pattern is set up in `CanvasElementProtocol` to listen for changes in element-specific properties.

Property changes are now correctly reflected on Firebase, but I'm having some trouble with displaying the changes once received.

There's also a weird issue where the elements in `canvasElements` have the latest `position`, `width`, `height`, and `rotation`, but the element-level objects do not. Hence during the observer notification the "new value" contains stale information. I've devised a workaround that uses the `position`, `width`, `height`, and `rotation` information from `canvasElements` instead, which works for now but we should probably investigate why it is happening and fix the root cause.

---

The received changes cannot be updated because each canvas element has its own view model, which has its own value-type model struct. This cannot be easily fixed, so this PR can be merged as-is for now.


ianyong (commented):
Deprecated by #61?

wltan:
Yes, I'll rework the collaboration feature

PR #57, opened 2021-04-04
Refactor uml connectors to a UmlConnectorView

hanstirtaputra:
Realized that when drawing the class diagram for Dynavity's View representation, UmlConnectors don't have their own View. As such I have added a `UmlConnectorView` to represent UmlConnectors, instead of using `Path`s in `CanvasElementMapView`


ianyong (approved):
LGTM!

PR #58, opened 2021-04-04
Fix unselecting a canvas element not clearing UML overlay

ianyong:
Previously, tapping on a plus button of `UMLSelectionOverlayView`, then deselecting the canvas element by tapping on the background of the canvas did not clear all of the plus buttons.


hanstirtaputra (approved):
LGTM!

PR #59, opened 2021-04-04
Defer quote autocorrect until runtime

wltan:
Previously when quotes such as `“` and `‘` are typed, they were previously immediately autocorrected to `"` and `'`. However, this has the unintended side effect of resetting the text cursor.

The autocorrection is now deferred until the code actually needs to be run.


ianyong (approved):
LGTM!

PR #64, opened 2021-04-08
(1) Persistence Layer (DTOs)

sebastiantoh:
This is one PR of a larger series of PRs that aims to implement the persistence layer.

The implementation in this PR is heavily inspired by https://betterprogramming.pub/why-model-objects-shouldnt-implement-swift-s-decodable-or-encodable-protocols-1249cb44d4b3 (open in incognito tab if you can't view the link)  with the use of Data Transfer Objects (this is basically the data representation of a model).

Other relevant links: 
- https://en.wikipedia.org/wiki/Data_mapper_pattern
- https://academy.realm.io/posts/isolating-your-data-layer/
- https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/infrastructure-persistence-layer-design
- https://martinfowler.com/eaaCatalog/repository.html

https://gist.github.com/omayib/9f515b6d5a72802bc2e07673788a308d is another link that might be useful to us. It's an example that has both a cloud storage as well as a cache, which has some similarities to what we are implementing.

In this PR, DTOs are added for most of our models (excluding UML-related stuff)

Other changes made:
- PDFElement now stores PDFDocument directly, rather than the URL

Stuff that I plan to do in the future:
- Add Repositories as per the link(s) above
- Add an interface for interacting with the local file system
- Remove conformance of `Codable` in all our models

Any feedback is welcome before I continue working on the stuff above!
- P.S. There's a lot of boilerplate introduced in this PR, and I'm actually not sure if this is ideal.


hanstirtaputra (dismissed):
LGTM!

> P.S. There's a lot of boilerplate introduced in this PR, and I'm actually not sure if this is ideal.

I think boilerplate here is fine as we are creating a separate layer for our persistence. Also, great articles, I think they're really useful!

Dynavity/Dynavity/persistence/data-mapping/CanvasDTO.swift
```diff
@@ -0,0 +1,23 @@
+struct CanvasDTO: Mappable {
+    typealias Model = Canvas
```
wltan: What do these aliases do? They don't seem to be used in the rest of the file, and the project seems to build properly even when commented out.

Edit: Just saw the associated type in `Mappable`. But it should be inferrable (for both the compiler and humans) what the type is, since each struct is responsible for only one type and does nothing else.

Dynavity/Dynavity/persistence/data-mapping/Mappable.swift
```diff
@@ -0,0 +1,5 @@
+protocol Mappable {
```
wltan, line 1: From what I can tell each DTO struct has a `Codable` extension. Does this requirement need to be captured in `Mappable`?

wltan (approved):
If you have nothing else to do on this PR, feel free to merge it so you don't have to maintain both this and #65

ianyong (approved):
LGTM! Thanks!

Dynavity/Dynavity/persistence/data-mapping/Mappable.swift
sebastiantoh: I don't think it should. `Mappable`s  should not necessarily be `Codable`s. It depends on how we want to store the underlying `Mappable`s. 
E.g. If we were to use a database, then it wouldn't be a `Codable`, but probably something else that can be converted into a database row

PR #65, opened 2021-04-09
(2) Persistence Layer (Local Storage) + App Integration

sebastiantoh:
Note: this PR is based off `persistence-layer`.

Changes:
- Remove `Codable` conformance in some models 
  - ~~The remaining models that still conform to Codable are mostly the enums and I'm not sure whether other functionalities depend on them being codable. Specifically, is there any reason why `CodeLanguage`, `UmlShape` and `UmlType` are `Int`s? Can they be`String`s instead?  @wltan @hansebastian )~~
  - Update: I've removed `Codable` conformance in all the models (there's still some commented out code relating to Codable which I haven't deleted yet in case we may need to reuse the implementation somewhere else)
- Add CanvasRepository (implemented using local storage)
  - Note: there is poor error handling here and this is something that can be looked into if we have time
-  Canvases will be automatically saved after some fixed time intervals, when the app loses focuses (e.g. app is minimized) and when the view disappears
    - These triggers should be good enough for now I think.
- Add checks in the UI to ensure that canvas names are unique
- All canvas renaming will only be done in the canvas selection screen (previously it was done in the sidemenu, this is no longer possible)
- Long press on the canvases in the selection view to bring up a context menu to delete / rename the individual canvas

~~Right now, canvases are stored in locally in the form `canvasName.json`. Canvases are assumed to have a unique name (the implementation of Hashable and Equatable assumes this). However, I'm thinking this will be a problem especially since we allow users to change the names of the canvases. E.g. there may be dangling `json` files which don't actually correspond to any canvases.~~
~~- One option is: everytime the user renames the canvas, I delete the old canvas and save another canvas with the new name~~
~~- Another option: save the file as `id.json`. I'm leaning towards this option, although this would bring up the issue of IDs again (which I think needs to be in the model)~~

Update: Canvases are stored as `id.json`, but the IDs are stored in the DTO. Names still need to be unique as the names are used to find the corresponding ID of the canvas

Additional stuff to be done
- Implement DTO for annotations and UML elements / connectors
- Integrate CloudStorage by conforming to `StorageManager`
- Saving of graph to local storage
- Improve error handling (if time permits)


wltan:
> Specifically, is there any reason why `CodeLanguage`, `UmlShape` and `UmlType` are `Int`s? Can they be `String`s instead?

Originally they were needed to conform to `Codable`. `String` should work just as fine, or if `Codable` is no longer required you should be able to drop the raw type altogether.

Dynavity/Dynavity/model/Todo.swift
```diff
@@ -1,6 +1,6 @@
 import SwiftUI
 
-struct Todo: Codable {
+struct Todo {
```
hanstirtaputra: These were not changed to classes, is it intentional?

EDIT: (Don't have to change it in this PR, just pointing it out)

hanstirtaputra (commented):
> Right now, canvases are stored in locally in the form canvasName.json. Canvases are assumed to have a unique name (the implementation of Hashable and Equatable assumes this). However, I'm thinking this will be a problem especially since we allow users to change the names of the canvases

Yep! When testing this PR, I encountered an issue where renaming the canvas will result in the app crashing as soon as it is being edited to a name that already exists.


https://user-images.githubusercontent.com/24553546/114266104-fda7cf00-9a26-11eb-99b9-5022d19666c4.mp4



> One option is: everytime the user renames the canvas, I delete the old canvas and save another canvas with the new name
Another option: save the file as id.json. I'm leaning towards this option, although this would bring up the issue of IDs again (which I think needs to be in the model)

I think this is a tough one, I'd say for the purpose of this project (and meeting with prof), it seems like the better choice is to not use IDs in the model? 

If the first option is really not ideal, could we instead have a mapping between the canvas name <--> ID, and keep this mapping and ID in the persistence layer? So on loading all canvases, we create this mapping, and maintain it whenever there are changes to the canvas name. Also, canvas names would have to be unique. Nonetheless, since name changes shouldn't occur often, I think the first option is acceptable.

sebastiantoh:
> Yep! When testing this PR, I encountered an issue where renaming the canvas will result in the app crashing as soon as it is being edited to a name that already exists.

Thanks for highlighting this issue! This is probably because of 2 canvases having the same hash value (due to having the same name). 

> If the first option is really not ideal, could we instead have a mapping between the canvas name <--> ID, and keep this mapping and ID in the persistence layer? So on loading all canvases, we create this mapping, and maintain it whenever there are changes to the canvas name. Also, canvas names would have to be unique. Nonetheless, since name changes shouldn't occur often, I think the first option is acceptable.

I'm not sure how this would work. The same issue you posted in the video would still occur right? 

I'm thinking of implementing checks in the UI instead to ensure that canvas names are unique and all canvas renaming will only be done in the canvas selection screen.
- Tap on new canvas button --> prompt for canvas name --> on clicking confirm, perform checks on canvas name--> optionally, go to canvas screen
- Long press a specific to bring up context menu --> options to rename / delete --> clicking rename will open a prompt for canvas name -> on clicking confirm, perform checks on canvas name

Dynavity/Dynavity/model/Todo.swift
sebastiantoh: Nope not intentional, I think it should be changed to classes and ID should probably be removed as well. @ianyong can confirm if this should be the case?

hanstirtaputra:
> I'm not sure how this would work. The same issue you posted in the video would still occur right?

Oh yes the issue in the video would still occur, sorry I wasn't clear I was trying to suggest an alternative for not storing ID's in model in response to the note you made that "Another option: save the file as id.json. I'm leaning towards this option, although this would bring up the issue of IDs again (which I think needs to be in the model)".

> I'm thinking of implementing checks in the UI instead to ensure that canvas names are unique and all canvas renaming will only be done in the canvas selection screen.
> 
> * Tap on new canvas button --> prompt for canvas name --> on clicking confirm, perform checks on canvas name--> optionally, go to canvas screen
> * Long press a specific to bring up context menu --> options to rename / delete --> clicking rename will open a prompt for canvas name -> on clicking confirm, perform checks on canvas name

Yep SGTM! I think anything that ensures uniqueness of names before saving the canvas would solve the issue? Can just implement whichever is simplest since UI is not our focus

sebastiantoh:
> Oh yes the issue in the video would still occur, sorry I wasn't clear I was trying to suggest an alternative for not storing ID's in model in response to the note you made that "Another option: save the file as id.json. I'm leaning towards this option, although this would bring up the issue of IDs again (which I think needs to be in the model)".

I see!

> Yep SGTM! I think anything that ensures uniqueness of names before saving the canvas would solve the issue? Can just implement whichever is simplest since UI is not our focus

Sure, I'll do this!

Update: I've also added an ID in the CanvasDTO since using the name as an identifier would mess up the graph if the canvases are renamed

Dynavity/Dynavity/model/Todo.swift
ianyong: Yea I think that would make sense. I can make the change if it isn't part of this PR.

sebastiantoh: I've made the change in this PR!

Dynavity/Dynavity/model/Todo.swift
```diff
@@ -1,10 +1,26 @@
 import SwiftUI
 
-struct Todo: Codable {
-    var id = UUID()
+class Todo {
     var label: String
-    var isCompleted = false
+    var isCompleted: Bool
```
ianyong, line 5: Toggling `isCompleted` no longer updates the view. I'll fix this in a separate PR to get this PR merged in as soon as possible.

Dynavity/Dynavity/model/Todo.swift
```diff
@@ -1,10 +1,26 @@
...
+    convenience init(label: String) {
+        self.init(label: label, isCompleted: false)
+    }
 }
 
 // MARK: Hashable
-extension Todo: Hashable {}
+extension Todo: Hashable {
+    static func == (lhs: Todo, rhs: Todo) -> Bool {
+        ObjectIdentifier(lhs) == ObjectIdentifier(rhs)
```
ianyong:
```suggestion
        lhs === rhs
```

Would this be better (to avoid the use of `ObjectIdentifier` unnecessarily)? This has exactly the same behaviour: https://github.com/apple/swift/blob/main/stdlib/public/core/Equatable.swift#L253.

Dynavity/Dynavity/persistence/storage/LocalStorageManager.swift
```diff
@@ -0,0 +1,93 @@
...
+        let paths = fileManager.urls(for: .documentDirectory, in: .userDomainMask)
+        return paths[0]
+    }
+
+    /// Returns the URL of the file in the documents directory
+    private func getFileURL(from name: String, ext: String) -> URL {
+        documentsDirectory.appendingPathComponent(name).appendingPathExtension(ext)
+    }
+
+    /// Checks if the file in the documents directory
```
ianyong:
```suggestion
    /// Checks if the file is in the documents directory
```

ianyong (dismissed):
LGTM! Thanks! Just some minor nits.

Dynavity/Dynavity/model/Todo.swift
sebastiantoh: Oops, sorry about that!

ianyong (approved):
LGTM! Thanks!

PR #66, opened 2021-04-09
Refactor UmlConnectors from struct to class

hanstirtaputra:
Known issues include connections drawing being off when `UmlElement` is rotated and resized. However I think this is not a big issue since it was mentioned that UI does not matter.

IMG_3C2B0335199B-1

### DIFF
--- a/Dynavity/Dynavity/view-model/canvas/CanvasViewModel+Uml.swift
+++ b/Dynavity/Dynavity/view-model/canvas/CanvasViewModel+Uml.swift
@@ -48,13 +48,11 @@ extension CanvasViewModel {
     }
 
     func updateUmlConnections(element: UmlElementProtocol) {
-        var idsToRemove: [UUID] = []
-        for var connector in canvas.umlConnectors {
+        for connector in canvas.umlConnectors {
             if connector.connects.fromElement !== element
                     && connector.connects.toElement !== element {
                 continue
             }
-            idsToRemove.append(connector.id)
             let sourceElement = connector.connects.fromElement
             let destElement = connector.connects.toElement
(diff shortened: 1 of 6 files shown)


Dynavity/Dynavity/view-model/canvas/CanvasViewModel+Uml.swift
```diff
@@ -96,11 +93,11 @@ extension CanvasViewModel {
 // MARK: UML connector gestures
 extension CanvasViewModel {
     func select(umlConnector: UmlConnector) {
-        if selectedUmlConnectorId == umlConnector.id {
-            selectedUmlConnectorId = nil
+        if selectedUmlConnector != nil {
```
sebastiantoh, line 96: Hmm, this is a different behaviour from what was previously implemented right? Previously it would only unselect if you tapped the same connector, now it will unselect even if you tap another connector, but not select the newly tapped connector

ianyong: This seems to be the case for canvas elements as well, so it's probably a regression I introduced. I'll fix this behaviour for both canvas elements and UML connectors in a separate PR. Thanks for pointing this out!

ianyong (approved):
LGTM!

Dynavity/Dynavity/view-model/canvas/CanvasViewModel+Uml.swift
ianyong: This seems to be the case for canvas elements as well, so it's probably a regression I introduced. I'll fix this behaviour for both canvas elements and UML connectors in a separate PR. Thanks for highlighting this issue!

PR #67, opened 2021-04-10
Fix behaviour when selecting canvas elements & UML connectors

ianyong:
Fixes the issue highlighted in https://github.com/Dynavity/dynavity/pull/66#discussion_r610989089.

Also updated the behaviour to enforce only one selection across all canvas elements and UML connectors. Previously, canvas element selections and UML connector selections were distinct.


hanstirtaputra (approved):
LGTM!

PR #70, opened 2021-04-11
Add persistence for UML elements and connectors

ianyong:
The mess in `UmlConnectorDTO` is so that we can store the relationships between UML connectors and UML elements without having to add an `id` field to the model (which Prof Wai Kay hates). This way, the IDs only exist in the DTO layer, leaving the model layer untouched.

Note that on initial load, the UML connectors are incorrectly offset, but this doesn't seem to stem from the persistence layer which this PR is about.

### DIFF
--- a/Dynavity/Dynavity/persistence/data-mapping/elements/uml-shapes/DiamondUmlElementDTO.swift
+++ b/Dynavity/Dynavity/persistence/data-mapping/elements/uml-shapes/DiamondUmlElementDTO.swift
@@ -9,7 +9,7 @@ struct DiamondUmlElementDTO: CanvasElementProtocolDTO, Mappable {
         self.canvasProperties = CanvasElementPropertiesDTO(model: model.canvasProperties)
         self.label = model.label
         self.umlType = model.umlType.rawValue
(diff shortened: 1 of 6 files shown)


Dynavity/Dynavity/persistence/data-mapping/elements/uml-shapes/DiamondUmlElementDTO.swift
```diff
@@ -0,0 +1,33 @@
...
+    let id: UUID
+
+    let canvasProperties: CanvasElementPropertiesDTO
+
+    let label: String
+    let umlType: String
+    let umlShape: String
+
+    init(model: DiamondUmlElement) {
+        self.id = UUID()
```
sebastiantoh, line 14: Does the id need to be the constant for the same connector across different reads / writes to the storage?

ianyong: Nope! If we wanted to do that though, we probably need to have it as part of the model?

sebastiantoh (approved):
LGTM!

PR #72, opened 2021-04-11
Fix annotation menu interfering with canvas element interactions

ianyong:
To test:
* Click on the pen mode twice to bring up the annotation menu.
* Click on the canvas element mode.
* Try to interact with any canvas element.


hanstirtaputra (approved):
LGTM!

PR #73, opened 2021-04-12
Add storage of backlinks

sebastiantoh:
Changes:
- Removed ID from BacklinkNode. Nodes must now be uniquely identified by their names
- Add storage of backlinks (adding of nodes, deleting of nodes, renaming of nodes)
  - Adding a node creates a node placed randomly in the graph
  - Deleting a node deletes the node and all connected edges
  - Renaming a node should preserve all edges
  - Adding / removing of edges can be done in the sidemenu of the respective canvases.
  - Longpressing on a node should go to corresponding canvas

### DIFF
--- a/Dynavity/Dynavity/DynavityApp.swift
+++ b/Dynavity/Dynavity/DynavityApp.swift
@@ -11,13 +11,11 @@ class AppDelegate: NSObject, UIApplicationDelegate {
 
 @main
 struct DynavityApp: App {
-    @StateObject var graphMapViewModel = GraphMapViewModel()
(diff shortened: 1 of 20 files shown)


hanstirtaputra (approved):
LGTM! Tested and works, cool to see this feature come together:)

Just one question, will there always only exist one backlink engine in the storage for the entire app?

sebastiantoh:
> Just one question, will there always only exist one backlink engine in the storage for the entire app?

For now, yes. Although if we were to implement what WK suggested of having multiple graph components to group related content together, I think we could have each component to be a separate backlink engine

PR #74, opened 2021-04-12
Add persistence for annotations

hanstirtaputra:
## Changes made:
1. Add annotations to local storage
2. Auto-save timer is increased from 3s to 10s

## Known issues:
1. Performance issues when there are many canvas elements and annotations. 
   - Saving to local storage causes the view to 'jitter'. For this reason I have increased the auto-save timer from 3s to 10s
2. Due to increase in auto-save timer, there are times where newly added annotations are not saved before navigating out of the `MainView`. (Saving on every update in annotations is not ideal either as it will affect the performance)

## Design decisions:
- Alternative 1: Store annotations in a separate `AnnotationsRepository`, and `AnnotationsDTO`
  -  Pros:
     - Separation of concerns between storage of canvas elements and annotations
     - Similar representation between model and storage
  - Cons:
     - Need to sync and maintain annotations and the canvases that they are linked to (during update, delete, read)
     - Require annotations to have their own unique ID for local storage (this can be done by using the name of the canvas) 
- Alternative 2: Store annotations together with `CanvasDTO`
  - Pros:
     - Simpler to sync between annotations and the canvas elements 
     - Need to create a new ID for annotations in the storage, since it is nested within the `CanvasDTO`
  - Cons:
     - `CanvasDTO` may be doing too many things
 
Decided to go with alternative 2 since we need not create a new ID just to store annotations. Furthermore, it makes our code less prone to synchronisation errors. Lastly, our annotations are technically a part of our canvas, it is just our design decision that our View separates the `Canvas` and `AnnotationCanvas` into two layers.


Dynavity/Dynavity/view/CanvasSelectionView.swift
```diff
@@ -224,4 +228,12 @@ extension CanvasSelectionView {
         errorAlert.addAction(UIAlertAction(title: "Dismiss", style: .cancel, handler: nil))
         self.showAlert(alert: errorAlert)
     }
+
+    private func unableToLoadAnnotationsHandler() {
```
sebastiantoh: This function isn't used anywhere right? Is this intentional?

Dynavity/Dynavity/persistence/data-mapping/CanvasWithAnnotation.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/**
+ Wrapper for Canvas and AnnotationCanvas to be stored together in the persistence layer.
+ */
+class CanvasWithAnnotation {
+    let canvas: Canvas
+    // Remove dependency of storage with PencilKit
+    let annotationCanvas: Data
```
sebastiantoh: Should this store the AnnotationCanvas directly instead to be more consistent with `Canvas`?
Also, should `CanvasWithAnnotation` be in the model? Otherwise the View-Model would be interacting with this directly, which is in the `Data Mapper` section

Dynavity/Dynavity/view-model/canvas/CanvasViewModel.swift
```diff
@@ -142,7 +144,10 @@ class CanvasViewModel: ObservableObject {
 // MARK: Local storage autosaving
 extension CanvasViewModel {
     func saveCanvas() {
-        self.canvasRepo.save(model: self.canvas)
+        let annotationData = self.annotationCanvas.drawing.dataRepresentation()
```
sebastiantoh: I think the `dataRepresentation()` part can be done when creating the canvas DTO

Dynavity/Dynavity/view/CanvasSelectionView.swift
hanstirtaputra: Nope not intentional, missed this one out, thanks!

Dynavity/Dynavity/persistence/data-mapping/CanvasWithAnnotation.swift
hanstirtaputra: Yep okay I think we can store `AnnotationCanvas` directly. Thanks!

>Also, should CanvasWithAnnotation be in the model? Otherwise the View-Model would be interacting with this directly, which is in the Data Mapper section

Yeap that makes sense too, but then it would mean that the `Data Mapper` section is interacting with the model?

sebastiantoh: > Yeap that makes sense too, but then it would mean that the Data Mapper section is interacting with the model?

Hmm what do you mean? You mean `CanvasDTO` would store a `AnnotationCanvas`? I was thinking `CanvasDTO` would store `Data` as it currently is right now.

hanstirtaputra: Nope I mean if we define the `CanvasWithAnnotation` class in the model, and we use this class in the `Data Mapper`, then that would mean that the `Data Mapper` is interacting with the model?

Perhaps I misunderstood your question on "Also, should CanvasWithAnnotation be in the model? Otherwise the View-Model would be interacting with this directly, which is in the Data Mapper section". My interpretation is that you're saying that the `CanvasWithAnnotation` class definition should be in the `model` directory instead of the `persistence` directory?

sebastiantoh: >  My interpretation is that you're saying that the CanvasWithAnnotation class definition should be in the model directory instead of the persistence directory?

Yep, this is what I meant!

> Nope I mean if we define the CanvasWithAnnotation class in the model, and we use this class in the Data Mapper, then that would mean that the Data Mapper is interacting with the model?

I think this is fine? All the other DTOs are sort of interacting with their corresponding model. So in this case, the mapping is between `CanvasDTO` <-> `CanvasWithAnnotation`

sebastiantoh: Also, now that I think about it, I'm not sure if we should store `Data` in the DTOs since that's quite `Codable`-specific. I think the DTOs should be independent of how we actually store the DTOs. But I think this is fine if there are no other alternatives

hanstirtaputra: Right makes sense, okay thanks!

sebastiantoh (approved):
LGTM, thanks!

PR #75, opened 2021-04-12
Add stronger validation on canvas names

sebastiantoh:
"Canvas name must be a unique, non-empty string, consisting of less than 20 alphanumeric characters or spaces. Canvas names are not case-sensitive."
- This is the conditions imposed on a canvas name and is also the error message that is shown to the users.
- (Is this error message too long?)


hanstirtaputra (approved):
LGTM! 

>(Is this error message too long?)

Yeah alert message is kinda long and might a little technical for non-tech/computing users, but I think its fine for this app's target audience!

PR #76, opened 2021-04-12
Implement cloud storage

wltan:
Allows the storage of canvases on Firebase, a partial reimplementation of #53.

1. Canvases always start in local storage upon creation.
2. Once a canvas is open, the canvas metadata/side menu now offers an option to publish the canvas to cloud storage instead.
3. Publishing a canvas removes it from local storage and adds it to cloud storage. Note that it is still tied to device ID, so canvases will by default not be visible on other devices.
4. If a canvas has been published, the side menu will display the unique canvas ID (including device ID). This will eventually be used for other devices to access this canvas (WIP).

## Major changes
- Added a new model object `OnlineCanvas` that extends from `Canvas`. This is necessary so that the app's various systems (e.g. view, viewmodel, storage) are able to differentiate local vs online canvases, and handle them accordingly.
  - The `OnlineCanvas` model *must* keep track of some identifier as it needs to be common across different devices (`ObjectIdentifier` fails in this regard)
- Added the corresponding DTO `OnlineCanvasDTO` that uses composition instead of inheritance.
- `StorageManager` has been made more abstract with the use of associated types, in order to support both `LocalStorageManager` and `CloudStorageManager` handling different DTOs.
  - This breaks the facade pattern, but it should not be a big problem for now.
  - If needed, new protocols extending from `StorageManager` could be added to specify the associated types, then other types conform to those protocols instead.
- `SideMenuView` now has access to `CanvasViewModel` using an environment object, instead of just the canvas name. This allows us to trigger the publication of the canvas.
- Removed the old Firebase implementation in `CanvasViewModel` as it is no longer required.

## Outstanding issues
- ~~While saving to cloud every few seconds appears fine, retrieving saved canvases is currently not working. There is a bug somewhere preventing the smooth conversion of Firebase data into `OnlineCanvasDTO` that I have yet to figure out.~~
  - Fixed. The problem is that Firebase does not store empty array values, thus received data will simply not contain all the necessary properties for decoding and fail. This is solved by writing a custom decoder for each DTO containing arrays to manually decode each property, substituting an empty array if the property is not present.
  - Currently only `CanvasDTO` and `TodoElementDTO` are affected as far as I'm aware, but this does not seem like a sustainable solution if more elements containing arrays are added to the model. We could explore a more general solution to this problem after STePS.
- The current implementation assumes `Canvas` contains all our necessary information instead of `CanvasWithAnnotation`. For functions that require the `CanvasWithAnnotation` type, some trivial wrapping and unwrapping is done. This will need to be cleaned up when `CanvasWithAnnotation` is removed and `AnnotationCanvas` becomes a part of `Canvas`.
  - This will be fixed in a separate PR.
- Sharing canvases to other devices and collaboration will be addressed in a separate PR.


Dynavity/Dynavity/persistence/data-mapping/OnlineCanvasDTO.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+struct OnlineCanvasDTO: Mappable {
+    let id: UUID
```
sebastiantoh: Is this field necessary, given that `CanvasDTO` already has an ID?

Dynavity/Dynavity/persistence/storage/CloudStorageManager.swift
```diff
@@ -0,0 +1,157 @@
+import Foundation
+import FirebaseDatabase
+import CodableFirebase
+import Combine
+
+struct CloudStorageManager {
+    let encoder = FirebaseEncoder()
+    let decoder = FirebaseDecoder()
+    let userId = OnlineCanvas.myUserId
```
sebastiantoh: Not really a big issue, but I'm not sure if the `storage` section should be reaching out to the model directly.

Dynavity/Dynavity/view/SideMenuView.swift
```diff
@@ -26,6 +29,15 @@ struct SideMenuView: View {
                 Text(canvasName)
                     .bold()
             }
+            if let onlineCanvas = canvasViewModel.canvas as? OnlineCanvas {
+                SideMenuContentView(label: "Shareable ID") {
+                    Text(onlineCanvas.shareableId)
+                }
+            } else {
+                Button("Publish") {
+                    sideMenuViewModel.publishCanvas(canvasViewModel: canvasViewModel)
```
sebastiantoh: Does this need to go through `SideMenuViewModel`? Can the `CanvasViewModel` publish directly?

Dynavity/Dynavity/model/canvas/OnlineCanvas.swift
```diff
@@ -0,0 +1,25 @@
...
+import SwiftUI
+
+class OnlineCanvas: Canvas {
+    static let myUserId = UIDevice.current.identifierForVendor?.uuidString
+            ?? "NO_USER_ID"
+
+    let id: UUID
+    let ownerId: String
+    var shareableId: String {
+        "\(ownerId)/\(id)"
```
sebastiantoh: Would it work if it was a concatenation with the canvas name instead of another ID? So that we don't explicitly store a UUID in the model

Dynavity/Dynavity/persistence/storage/StorageManager.swift
```diff
@@ -1,7 +1,9 @@
...
-    func saveCanvas(canvas: CanvasDTO) throws
-    func deleteCanvas(canvas: CanvasDTO) throws
+    associatedtype ModelDTO
+
+    func readAll() throws -> [ModelDTO]
+    func save(model: ModelDTO) throws
+    func delete(model: ModelDTO) throws
 
     func readBacklinkEngine() throws -> BacklinkEngineDTO?
     func saveBacklinkEngine(backlinkEngine: BacklinkEngineDTO) throws
```
sebastiantoh, line 11: Would it be better if CloudStorageManager conforms to this protocol as well, but fatalErrors for these methods?

Dynavity/Dynavity/persistence/storage/CloudStorageManager.swift
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+// MARK: Struct to help with decoding
+struct ExternalCanvasReference: Codable, Equatable {
+    let userId: String
+    let canvasId: String
+}
+
+// MARK: Helper to synchronize asynchronous code
```
wltan: This converts all the asynchronous stuff from Firebase into synchronous code by blocking until a value is received.

It can move to its own package once the implementation is more stable.

Dynavity/Dynavity/persistence/data-mapping/OnlineCanvasDTO.swift
wltan: Not really.

I've cleaned up the storage manager to reduce the amount of redundant data on Firebase.

Dynavity/Dynavity/view/SideMenuView.swift
wltan: My original intention was to allow `SideMenuViewModel` to make any additional actions if necessary instead of interacting directly with another view-model, which can create coupling across different view/view-models.

But the logic turns out to be quite trivial and I see that `GraphViewModel` is being directly accessed too, so I guess publishing directly should be ok.

Dynavity/Dynavity/persistence/storage/StorageManager.swift
wltan: I think I might have overlooked the conformance when fixing merge conflicts yesterday.

To get around the differing requirements, I made two new protocols (extending the original generic one) and have each manager conform to those instead.

Dynavity/Dynavity/model/canvas/OnlineCanvas.swift
wltan: This should work in theory but by that logic ID should no longer be necessary in storage either (including `CanvasDTO`), since canvas name is consistent within a device and is unique globally when combined with owner ID.

Significantly more work will be needed to remove canvas IDs (again), which should probably be done in another PR.

Dynavity/Dynavity/persistence/storage/CloudStorageManager.swift
wltan: I've changed it to go through `OnlineCanvasDTO` instead of querying `OnlineCanvas` directly, since we are dealing with `OnlineCanvasDTO` more. That should help?

Dynavity/Dynavity/model/canvas/OnlineCanvas.swift
sebastiantoh: Hmm the reason why there's an ID in CanvasDTO is to support renaming of canvases. I'm not sure if it would work if we remove ID from CanvasDTO

> Significantly more work will be needed to remove canvas IDs (again), which should probably be done in another PR.

Yep, ok!

wltan: Does this approach not work? https://github.com/Dynavity/dynavity/pull/76/files#diff-860904f80aae08fea7cc64af7e8226cb3d4ce2a70ef18359ef0207c3bbc16059R52

I rewrote the replace function to simply combine a delete and save operation. This allows it to select the appropriate methods based on whether the canvas is an `OnlineCanvas` or not.

sebastiantoh: Oh, right. That should work

sebastiantoh:
After pressing `Publish`, I should still be able to see the published canvas in the selection screen right?

wltan:
> After pressing `Publish`, I should still be able to see the published canvas in the selection screen right?

Yes, but it no longer exists in local storage so what you see on the selection screen is from the cloud storage.

sebastiantoh (approved):
LGTM!

There's still some issues that I'm facing, not sure if it's only me. Maybe @ianyong @hansebastian can test and verify as well.

Unable to write 'large' canvases to FireBase:
- Can try reproducing by loading the sample canvases I sent and publishing the `Sorting` canvas (294KiB)
image

Occassional unresponsiveness
- I'm guessing this is due to the blocking reads / writes 
- Or maybe due to the CPU spikes even when idling (this is not present on the current `master`). CPU spikes seem to line up with network requests (either before or after a request is made)
image

hanstirtaputra:
> There's still some issues that I'm facing, not sure if it's only me. Maybe @ianyong @hansebastian can test and verify as well.
> Unable to write 'large' canvases to FireBase:

Yep I've tested it on my end for `Sorting` canvas and can confirm that I encountered a `write_too_big` failure as well

wltan:
This might be a limitation of Firebase: https://firebase.google.com/docs/database/usage/limits
Perhaps we can just document it as such?

I'll merge this first to continue on the collaboration feature. If there is time we can look into more efficient representations of a canvas, which hopefully solves both issues stated above.

PR #79, opened 2021-04-13
More fixes

sebastiantoh:
Sorry for the spam.

- Forgot to update one more part regarding backlink storage file naming
- Update default mode to be selection
- Update hashable impl. of CanvasWithAnnotation

### DIFF
--- a/Dynavity/Dynavity/model/canvas/CanvasWithAnnotation.swift
+++ b/Dynavity/Dynavity/model/canvas/CanvasWithAnnotation.swift
@@ -22,6 +22,6 @@ extension CanvasWithAnnotation: Equatable, Hashable {
     }
 
     public func hash(into hasher: inout Hasher) {
-        hasher.combine(ObjectIdentifier(self).hashValue)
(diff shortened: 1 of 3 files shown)


Dynavity/Dynavity/model/canvas/CanvasWithAnnotation.swift
```diff
@@ -22,6 +22,6 @@ extension CanvasWithAnnotation: Equatable, Hashable {
     }
 
     public func hash(into hasher: inout Hasher) {
-        hasher.combine(ObjectIdentifier(self).hashValue)
+        hasher.combine(self.name)
```
sebastiantoh, line 25: @hansebastian This change is OK right? I made this change because

```
Two instances that are equal must feed the same values to Hasher in hash(into:), in the same order.
```

hanstirtaputra: Yep this is okay!

hanstirtaputra (approved):
LGTM!

PR #80, opened 2021-04-13
Polish keyboard behaviour

ianyong:
Previously, whenever the keyboard appears, the `CanvasElementMapView` would be shifted upwards, but not the `AnnotationCanvasView`s. This caused the annotations to be offset from the canvas elements. This PR disables the keyboard avoidance behaviour such that the `CanvasElementMapView` no longer shifts upwards when the keyboard is opened. With or without the keyboard avoidance behaviour, there will always be instances where the keyboard blocks the view of whatever is being typed.

In addition, the keyboard is now closed if the user taps on the background of the canvas, or opens the UML diagram menu.


wltan (approved):
Works as described above

PR #81, opened 2021-04-13
Fix canvas elements & annotations being out of sync

ianyong:
No idea why this is the case. I spent so long debugging my calculations before I realised that the problem was due to the wrong viewport size being reported.


sebastiantoh (approved):
LGTM!

PR #82, opened 2021-04-13
Unify annotations under Canvas

wltan:
This PR builds off the existing branch for #76, **and targets that branch instead of master**. Otherwise, the changes made are difficult to review and there will be merge conflicts with master.

`Canvas` now contains `AnnotationCanvas` as a property, so that any component that wishes to know everything about a canvas can do so through `Canvas` and not `CanvasWithAnnotation` (which is now removed).

## Issues

Canvas elements and annotations appear to be slightly bugged when repeatedly navigating between the canvas selection screen and the canvas view. This same behavior exists in the master branch too and is unlikely to be caused by these changes.


sebastiantoh (approved):
LGTM!

> Canvas elements and annotations appear to be slightly bugged when repeatedly navigating between the canvas selection screen and the canvas view. This same behavior exists in the master branch too and is unlikely to be caused by these changes.

Yep, thanks for raising this!

PR #83, opened 2021-04-14
Implement canvas-level collaboration

wltan:
This allows multiple devices to work on a common canvas.
1. On the canvas selection screen, there is a new icon next to the New Canvas button. This allows the user to import an already shared canvas on the cloud using the shared ID (implemented in #76).
  a. There is currently no way to efficiently copy the shared ID from the app (e.g. to share it with others). This is a QoL issue that can easily be fixed in future. 
  b. The imported canvas is represented on the backend as a "foreign key" of sorts, referring to the canvas stored under the original creator.
3. Any device will be able to work on a canvas as if though it were owned by that user. Writes will be visible to all users who either own or have imported that canvas.
4. The latest state is currently retrieved **once** from the backend **upon launch**, as is the case for all canvases.
  a. Note that working simultaneously on the same canvas (aka google docs behavior) is **not implemented yet**. If multiple devices are open with the same canvas, they will conflict with each other and cause undefined behavior.

### DIFF
--- a/Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
+++ b/Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
@@ -93,6 +93,12 @@ struct CanvasRepository: Repository {
     }
 
     func importCanvas(ownerId: String, canvasName: String) -> Bool {
-        false
+        do {
+            let canvas = try cloudStorageManager.importCanvas(ownerId: ownerId, canvasName: canvasName)
+            return canvas != nil
+        } catch {
+            return false
+        }
+
     }
 }
--- a/Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
+++ b/Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
@@ -91,4 +91,8 @@ struct CanvasRepository: Repository {
         try? localStorageManager.readAll()
             .first(where: { $0.name == canvas.name })?.id
     }
+
+    func importCanvas(ownerId: String, canvasName: String) -> Bool {
+        false
+    }
 }
--- a/Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
+++ b/Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
@@ -59,7 +59,7 @@ struct CanvasRepository: Repository {
         }
     }
 
-    func deleteLocal(model: Canvas) -> Bool {
+    private func deleteLocal(model: Canvas) -> Bool {
         let id = getExistingId(for: model) ?? UUID()
         let canvasDTO = CanvasDTO(id: id, model: model)
         do {
@@ -70,7 +70,7 @@ struct CanvasRepository: Repository {
         }
     }
 
-    func deleteCloud(model: OnlineCanvas) -> Bool {
+    private func deleteCloud(model: OnlineCanvas) -> Bool {
         do {
             try cloudStorageManager.delete(model: OnlineCanvasDTO(model: model))
             return true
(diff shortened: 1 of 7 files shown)


Dynavity/Dynavity/persistence/repositories/CanvasRepository.swift
```diff
@@ -91,4 +91,14 @@ struct CanvasRepository: Repository {
...
     }
+
+    func importCanvas(ownerId: String, canvasName: String) -> Bool {
+        do {
+            let canvas = try cloudStorageManager.importCanvas(ownerId: ownerId, canvasName: canvasName)
+            return canvas != nil
+        } catch {
+            return false
+        }
+
```
ianyong:
```suggestion
```

ianyong (dismissed):
Not able to import a shared canvas due to restrictions on the input. In particular, I am testing with the following canvas ID: `A30CAE07-3312-4D04-9714-E17471E90E85/test`.

Dynavity/Dynavity/view/CanvasSelectionView.swift
```diff
@@ -225,13 +230,62 @@ extension CanvasSelectionView {
         self.showAlert(alert: alert)
     }
 
+    private func onImportCanvasButtonTap() {
+        let alert = UIAlertController(title: "Import shared canvas", message: nil, preferredStyle: .alert)
+        alert.addTextField { textField in
+            textField.placeholder = "Enter shareable ID here"
+        }
+        alert.addAction(UIAlertAction(title: "Save", style: .default, handler: { _ in
```
sebastiantoh:
```suggestion
        alert.addAction(UIAlertAction(title: "Import", style: .default, handler: { _ in
```

Dynavity/Dynavity/view-model/CanvasSelectionViewModel.swift
```diff
@@ -57,6 +57,10 @@ class CanvasSelectionViewModel: ObservableObject {
         self.objectWillChange.send()
     }
 
+    func importCanvas(name: String, owner: String) -> Bool {
+        canvasRepo.importCanvas(ownerId: owner, canvasName: name)
```
sebastiantoh: Otherwise, the UI will not update after importing
```suggestion
        canvasRepo.importCanvas(ownerId: owner, canvasName: name)
        self.objectWillChange.send()
```

Can help to add the above change to `CanvasViewModel`'s `saveCanvas` as well?

wltan:
> Not able to import a shared canvas due to restrictions on the input. In particular, I am testing with the following canvas ID: `A30CAE07-3312-4D04-9714-E17471E90E85/test`.

Seems to work fine on my end. What error are you getting?

ianyong:
> > Not able to import a shared canvas due to restrictions on the input. In particular, I am testing with the following canvas ID: `A30CAE07-3312-4D04-9714-E17471E90E85/test`.
> 
> Seems to work fine on my end. What error are you getting?

Hmm, I'm getting the following error:
image

sebastiantoh:
> Hmm, I'm getting the following error:
> image

I think you can't import another canvas named `test` if you already have another canvas named `test`

wltan:
> I think you can't import another canvas named `test` if you already have another canvas named `test`

Right, perhaps that could be made clearer in the error message.

I've extracted out a `showErrorAlert` function that we can use to call the error alert more frequently with more specific messages.

ianyong:
> > Hmm, I'm getting the following error:
> > image
> 
> I think you can't import another canvas named `test` if you already have another canvas named `test`

Ah I see, thanks! How did you all test the collaboration then?

sebastiantoh:
> I've extracted out a `showErrorAlert` function that we can use to call the error alert more frequently with more specific messages.

Yep, I'll improve the error messages if there's nothing urgent to be done


> Ah I see, thanks! How did you all test the collaboration then?

 I just went to the firebase console and selected a random canvas to import

sebastiantoh (approved):
LGTM

PR #84, opened 2021-04-15
Improve error messages

sebastiantoh:
Error messages are now more specific as opposed to a generic error message.

Context: https://github.com/Dynavity/dynavity/pull/83#issuecomment-819254234


ianyong (approved):
LGTM! Tangentially, there appears to be an issue when renaming a canvas that you just exited out of (a new canvas with the old name will pop up after a while).

PR #85, opened 2021-04-17
Clean up code

ianyong:


wltan:
For the todos that were removed, are we no longer planning to work on those?

Dynavity/Dynavity/graph/Graph.swift
```diff
@@ -248,9 +248,6 @@ struct Graph<T: Hashable> {
 
     /// Checks the representation invariants.
     private func checkRepresentation() -> Bool {
-        // TODO: remember to disable this / ship app in production
-        // The next line is here for ease of commenting / uncommenting this function out
-        true ||
```
ianyong, line 253: Since `checkRepresentation` is only ever called in assertions, we just need to build the app in release mode.

Dynavity/Dynavity/view/UmlSideMenuView.swift
```diff
@@ -6,7 +6,6 @@ struct UmlSideMenuView: View {
     private let umlCloseButtonOffset: CGFloat = -12.0
     private let shapePreviewSize: CGFloat = 80.0
 
-    // TODO: Move this to the model, so that view will always be in sync
```
ianyong, line 9: I don't think that this should be in the model since this is effectively just a list of UML elements which should appear in the side menu. Either way, to get a list of types which conform to `UmlElementProtocol` automatically, we would probably need to use some form of type reflection (such as https://gist.github.com/codelynx/bb72bf0bed58a327ce1dbe6639d0369b) which is a bit iffy.

Dynavity/Dynavity/view/canvas/CanvasThumbnailView.swift
```diff
@@ -5,7 +5,6 @@ struct CanvasThumbnailView: View {
     var isSelected: Bool
     var body: some View {
         VStack {
-            // TODO: replace this with canvas thumbnail
```
ianyong, line 8: This is a UI only change so it is not a priority.

Dynavity/Dynavity/view/canvas/PKCanvasWrapperView.swift
```diff
@@ -9,6 +9,5 @@ class PKCanvasWrapperView: PKCanvasView {
     override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
         // This maintains the inherited functionality of PKCanvasView to do the drawing
         super.touchesBegan(touches, with: event)
-        // TODO: Possibly add CanvasElement on user's touch location
```
ianyong, line 12: Canvas elements are added to the center of the screen since #49.

ianyong (commented):
Added reasoning for the removal of all the TODOs. As for the `DynavityUITests` target and `DynavityTests.swift` file, those are unused and contain only boilerplate code.

Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
```diff
@@ -3,7 +3,7 @@ import PencilKit
 
 struct AnnotationCanvasView: View {
     @ObservedObject var viewModel: CanvasViewModel
-    @State private var annotationCanvasView = PKCanvasWrapperView()
+    @State private var annotationCanvasView = PKCanvasView()
```
sebastiantoh, line 6: Just checking, does this actually need to be a `@State`? `AnnotationCanvasView` itself is essentially a `PKCanvasView` right?

sebastiantoh (dismissed):
LGTM

Dynavity/Dynavity/view/canvas/AnnotationCanvasView.swift
ianyong: Yup, it does. If the `@State` property wrapper were to be removed, then the `AnnotationCanvasView` would not invalidate its appearance and recompute its body when the `annotationCanvasView` is updated. This can be observed by drawing some annotation onto the canvas, switching to the canvas element mode, then zooming/scrolling. The `AnnotationCanvasView` that is responsible for displaying the annotations will not update as the view is not automatically invalidating itself when the `AnnotationCanvasView` receives updates from the `CanvasViewModel`.

ianyong: With regards to your second question, `AnnotationCanvasView` is a wrapper for `PKCanvasView`. Hence, the removal of `PKCanvasWrapperView` as there is no point having two layers of wrappers.

sebastiantoh: From my understanding of how `UIViewRepresentable` works, SwiftUI will call the `makeUIView` method, which returns a `PKCanvasView`. So `AnnotationCanvasView` and `PKCanvasView` are essentially the same? 

Edit: nvm, I realised that `updateUIView` updates some other view, rather than the UIView itself (the parameter is unused)

sebastiantoh (approved):
LGTM!

PR #88, opened 2021-04-17
Fix uml element deletion

hanstirtaputra:
Deleting UML elements now delete the `UmlConnector`s that are connected to it. This fixes the issue where loading from storage fails when `UmlConnector`s can not find the corresponding UML elements that were deleted.


ianyong (approved):
LGTM!

PR #89, opened 2021-04-17
Implement live collaboration

wltan:
This allows canvas data to be updated in real-time when changes are made to the backend data (e.g. by another device or manual tampering on the Firebase console). Four properties of a canvas are to be synchronized:
1. List of canvas elements
2. Annotation canvas
3. List of UML elements
4. List of UML connectors (not working)

Note that Firebase treats arrays as atomic values. This means that our lists of elements are saved and written wholesale, and there is no convenient way to observe each element individually without an overhaul of the storage representation.

## Issues
- Although pulling changes is instantaneous, there is still a delay in writing changes to Firebase (which is defined by how often the canvas autosaves). This still creates race conditions where the first device to save to Firebase overwrites all other devices, regardless if changes have been made or not.
  - This can be solved by either making Firebase save instantly when changes are made, or perhaps prevent saving if no changes are made.
- [Fixed] UML connectors are unable to be synchronized. This is because connectors are stored with a pair of IDs describing the two elements it connects, but the mapping of IDs to elements is lost after the initial canvas loading.
  - This has been fixed by combining UML elements and connectors under a single key in `CanvasDTO`, so by listening for changes on that key, we will always see updates for connectors together with the corresponding list of elements.
- [Fixed] When changes to the annotation canvas are received, the value in `Canvas` is updated but the changes are not propagated to the view.


Dynavity/Dynavity/persistence/data-mapping/OnlineCanvasDTO.swift
```diff
@@ -17,8 +17,8 @@ struct OnlineCanvasDTO: Mappable {
     }
 
     func toModel() -> OnlineCanvas {
-        OnlineCanvas(ownerId: ownerId, canvas: canvas.toModel())
+        let model = OnlineCanvas(ownerId: ownerId, canvas: canvas.toModel())
+        CanvasRepository().addCloudChangeListeners(model: model)
```
sebastiantoh, line 21: Would it be better to have a wrapper method in `CanvasRepository`  that calls `toModel` of the `OnlineCanvasDTO`, then adds the listeners? This is to remove the dependency from DTO to Repository.
Or maybe `addCloudChangeListeners` in `CanvasRepository` can be another method that takes in an `OnlineCanvasDTO`, converts it to a model, adds the listeners, then returns the `OnlineCanvas`?

Dynavity/Dynavity/persistence/storage/CloudStorageManager.swift
```diff
@@ -135,6 +136,40 @@ struct CloudStorageManager: OnlineStorageManager {
         // return imported canvas
         return canvas
     }
+
+    func addChangeListeners(model: OnlineCanvas) {
```
sebastiantoh, line 140: Can this take in an `OnlineCanvasDTO` and perform the changes on the DTO instead?
So that the actual persistence implementation does not know about the model

Dynavity/Dynavity/model/canvas/Canvas.swift
```diff
@@ -36,25 +40,34 @@ class Canvas: ObservableObject {
...
 
-        for connector in umlConnectors {
-            if connector.connects.fromElement !== umlElement
-                    && connector.connects.toElement !== umlElement {
-                continue
-            }
-            removeUmlConnector(connector)
-        }
+    func replace(annotation: AnnotationCanvas) {
+        self.annotationCanvas = annotation
```
hanstirtaputra, line 58: ~Perhaps the reason why the annotations changes are not being reflected is because the entire `AnnotationCanvas` is replaced, thus the observers don't work? Could we instead replace the `drawing` within the `annotationCanvas` instead of replacing the entire `annotationCanvas` with a new one?~

EDIT: The issue has been addressed by #90

hanstirtaputra (commented):
> UML connectors are unable to be synchronized. This is because connectors are stored with a pair of IDs describing the two elements it connects, but the mapping of IDs to elements is lost after the initial canvas loading.
This can be fixed by either maintaining an ID in the model, or retrieving the canvas elements again to obtain their identifiers.

I think its preferable to retrieve the canvas elements since prof does not like having IDs in models. To add on, I think an alternative where the `UmlConnector`s does not use IDs is not possible since class references are lost when converting from model to storage?

Dynavity/Dynavity/persistence/data-mapping/OnlineCanvasDTO.swift
wltan: `addCloudChangeListeners` is not exclusively used in `OnlineCanvasDTO`. When a canvas is published, a new `OnlineCanvas` is made, and that will also require listeners to be attached.

I do admit however that there is a bit of duplication going on, since in principle we want every `OnlineCanvas` to be listening to the cloud (or whatever `OnlineStorageManager` the `CanvasRepository` has). I did consider performing `addCloudChangeListeners` in the OnlineCanvas initialiser, but that might open another bunch of problems.

Dynavity/Dynavity/persistence/storage/CloudStorageManager.swift
wltan: I don't think this will work since the purpose of change listeners is to directly update the `OnlineCanvas` state that is used by `CanvasViewModel`. The `OnlineCanvasDTO` will not be able to do this as it has no association to the model once `toModel` finishes.

An alternative (instead of calling `OnlineCanvasDTO` -> `CanvasRepository` -> `CloudStorageManager`) is to try and set up the listeners when `CloudStorageManager` initialises `OnlineCanvasDTO`. However, in this approach all the observer setup needs to be deferred until the model object is created, using copious amounts of escaping closures.

hanstirtaputra (approved):
LGTM!

PR #92, opened 2021-04-18
Remove use of NavigationLink

sebastiantoh:
This should improve the memory not being freed from repeatedly navigating between a canvas and the homescreen


ianyong (approved):
LGTM! Checked against the profiler and memory graph. :tada:

